Issue #57, opened 2018-07-05
Implicitly discover schema and queries?


tomhoule:
Schema path: based on .graphqlconfig

Queries: expect a file named like the struct next to the rust file.

This was suggested by @brainlessdeveloper

This would be a fallback if the explicit paths are omitted of course.

tomhoule:
@fnune I think we can close this issue :p

fnune:
😁 Congrats on growing this into what it is now!

tomhoule:
Well, I had no bandwidth to maintain this for a long time, so the crate has been stagnant. Would have love for someone else to take over. But now I'm getting back to maintaining, I have a bunch of ideas.

Issue #31, opened 2018-06-25
Support multiple operations per query document


tomhoule:
Probably via a struct attribute on the query struct to specify which operation it should implement.

For now we should document that we only support one operation by document (like Apollo Client, I think?)

tomhoule:
This is important because it's the most natural way to share fragments between multiple operations (an operation is either a query, a mutation or a subscription).

tomhoule:
I started implementing that - the question is whether we should keep the 1:1 match between a struct under derive and an operation, or generate multiple implementors of the `GraphQLQuery` trait (one for each operation).

I tend to prefer the first option, as it's more explicit and gives more control to users (and it's not a lot of boilerplate to define additional structs with nearly the same options but a different name).

tomhoule:
Question: should we select the operation with the name of the struct by default, or the first one, or a combination of both? I think the struct name, with a fallback on the first one makes sense - or maybe struct name, and error if there is no match.

tomhoule:
done!

PR #60, opened 2018-07-11
WIP - Implicitly discover schema and query without macro attributes

fnune:
So, this is the most naïve implementation I could fathom. Current problems:

1. `to_snake_case` won't take care of numbers. In the existing example, `Query1` is snake-cased as `query1` but the file name was `query_1`. I suppose this is not a big issue.
2. `implicitly_discover_schema_name` is extremely verbose, I suppose you can come up with some nicer way of writing this.
3. This assumes there's `src/`. There should probably be another approach, and I'd like to store those magic strings somewhere nicer.
4. I'm probably using some archaic way of manipulating strings.


brainlessdeveloper:
https://github.com/tomhoule/graphql-client/issues/57

graphql_query_derive/src/lib.rs
```diff
@@ -66,6 +66,26 @@ fn read_file(
...
+fn implicitly_discover_schema_name(cargo_manifest_dir: &str) -> Result<String, failure::Error> {
+    if ::std::path::Path::new(&cargo_manifest_dir).join("src/schema.json").exists() {
+        return Ok("src/schema.json".to_owned());
+    }
+    if ::std::path::Path::new(&cargo_manifest_dir).join("src/schema.graphql").exists() {
+        return Ok("src/schema.graphql".to_owned());
+    }
+    if ::std::path::Path::new(&cargo_manifest_dir).join("src/schema.gql").exists() {
+        return Ok("src/schema.gql".to_owned());
+    }
```
brainlessdeveloper: I'm sure there's a much more elegant way of writing this, but I'm a newb so I'll gladly take suggestions.

scrogson:
2. How about `infer_schema_name`?

graphql_query_derive/src/lib.rs
scrogson: What about something like this?

```rust
fn find_schema_file(cargo_manifest_dir: &str) -> Result<String, failure::Error> {
    let default_paths = vec!["src/schema.json", "src/schema.graphql", "src/schema.gql"];

    default_paths
        .iter()
        .map(|path| Path::new(&cargo_manifest_dir).join(path))
        .find(|path| path.exists())
        .and_then(|path| path.to_owned())
        .ok_or_else(|| Err(format_err!("None of the following exist in src/: schema.json, schema.graphql, schema.gql")))
}
```

brainlessdeveloper: That looks much nicer! Thanks

brainlessdeveloper:
I just realized I should be parsing `.graphql` for the schema path.

scrogson:
> I just realized I should be parsing .graphql for the schema path.

Do you mean `. graphqlconfig`?

graphql_query_derive/src/lib.rs
tomhoule: +1 for this solution. I'm not sure but I think it can even be an array literal instead of `vec!`, so `let paths = &["src/schema.json", "src/schema.graphql", "src/schema.gql"];`, which would be a tiny bit more efficient but we really don't care in that context since it only runs once per query at compile-time.

tomhoule:
Really nice, thanks for the PR! There are two things I'm not sure about myself, so it's open for discussion:

1. Is `src/schema.{json,graphql}` a reasonable default? In the issue I mentioned maybe using a schema from `.graphqlconfig` since it is a standard for the JS GraphQL tooling. Having the schema at the project root, next to the Cargo.toml could also be a reasonable default.
2. Should we make a separate test for that (with a tiny query and schema since we only want to check that the files are found) in the `tests/` dir or just use the github examlpe as a test?

For queries I think it's fine this way, we shouldn't worry about case conversion edge-cases since there's always the explicit `query_path` option as a fallback.

tomhoule:
see https://github.com/prismagraphql/graphql-config for `.graphqlconfig`

graphql_query_derive/src/lib.rs
```diff
@@ -75,8 +95,14 @@ fn impl_gql_query(input: &syn::DeriveInput) -> Result<TokenStream, failure::Erro
     let cargo_manifest_dir =
         ::std::env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR env variable is defined");
 
-    let query_path = extract_attr(input, "query_path")?;
-    let schema_path = extract_attr(input, "schema_path")?;
+    let module_name = Ident::new(&input.ident.to_string().to_snake_case(), Span::call_site());
+    let struct_name = &input.ident;
+
+    let query_path = extract_attr(input, "query_path")
+        .or_else(|_err| implicitly_discover_query_name(&struct_name.to_string()))?;
```
tomhoule: There's an error message line 60 in this file that gets printed if the file isn't found, we should update it with info on how we try to find the file explicitly, otherwise it will be confusing. The problem is that in `read_file` we don't know whether the path was specified implicitly or just inferred so we can't adapt the error message. Do you have an idea? I'll try to think of something too.

graphql_query_derive/src/lib.rs
scrogson: Yeah, I believe that should work as well. 👍

scrogson:
> Having the schema at the project root, next to the Cargo.toml could also be a reasonable default.

I think I agree with this.

brainlessdeveloper:
So for the schema we could follow something like this:

0. Use the macro attribute.
1. Try to parse `.graphqlconfig` and find it from there (maybe it's worth trying to have a file dedicated to handling stuff from `.graphqlconfig`?).
2. If it's not there, try to find it in the project root next to `Cargo.toml`.
3. Otherwise try to find it in `${root}/src/`.
4. Fail. 

For the query files, I'm thinking defaulting to `${root}/src/` is not the best idea, because most projects will have query files somewhere else in the code. Maybe (if not too magic) something like this would work:

0. Use the macro attribute.
1. Try to find the file searching the directory from which the macro was called, and recurse parent directories until the project root. Maybe also take into account `includes` and `excludes` from `.graphqlconfig`? Not sure if this would be worth it.
2. Fail.

In any case, we could just start by using the defaults you've suggested. What do you think?

tomhoule:
For the schema I agree, as long as we properly document the implicit behaviour so it doesn't feel too magic.

For the queries I was thinking we should look for the query file next to the file where the struct is located (same directory)  but it turns out it's not easy. I am wondering about another feature we want (multiple operations per query document: https://github.com/tomhoule/graphql-client/issues/31). We could use the struct name to easily select an operation inside a query file (as a fallback from an explicit setting or finding the single query in the file) instead, and now it seems more natural to me. What do you think?

tomhoule:
For `.graphqlconfig` yes it's better not to do that in `lib.rs`. But maybe it can be done in another PR if that's too much for this one (I don't know how much work that could be).

brainlessdeveloper:
Updates:

https://crates.io/crates/graphql-config
https://github.com/brainlessdeveloper/graphql-config

brainlessdeveloper:
Does it still make sense to pursue this after so much development has happened on master?

tomhoule:
It would be hard to rebase I think. And nobody has requested the feature so far, so I don't know if it should be pursued. Sorry :/

tomhoule:
@fnune the library changed _a lot_ since this PR was open, I think we should start from scratch if we still want to do something like this. I'm working on graphql-client a bit again, so we can get ideas going in issues. I have a few for a better codegen workflow.

fnune:
Cool! What are the plans?

tomhoule:
I want to completely decouple `graphql_client_codegen` from the compiler's macro architecture, since code generation is based on the GraphQL query + schema, not rust code. That would make it easy to have an in-browser playground that shows you exactly what gets generated.

For the "auto-discovery", the idea would rather be a "batch generation" macro that generates all the modules for your query at once, so you don't need to write the paths of query files as often.

Issue #83, opened 2018-08-22
How to use the CLI client to generate a schema?


djc:
I've built a Juniper-based GraphQL server this week and now I'd like to execute tests against it. This project seems like the best way to do so from Rust, and so I'm trying to generate a schema for my server. The built-in GraphiQL instance can read the introspection data just fine, but when I run `graphql-client introspect-schema http://localhost:3000/graphql` I just get `{"data":{"schema":null},"errors":null}`.

(Just to be sure, I get the same result if I run `introspect-schema` against the GitHub GraphQL endpoint.)

What, if anything, am I getting wrong?

tomhoule:
I haven't paid attention to the CLI since its initial release nearly two months ago so it may have broken. I will definitely investigate this asap but as a temporary solution you can use https://github.com/apollographql/apollo-cli for introspecting and downloading the schema.json

djc:
I have been investigating, but can't quite figure it out. As far as I can tell the problem is in the deserialization of the introspection query result -- can you recommend any techniques for debugging the deserialization code that runs during `GraphQLResponse` deserialization?

tomhoule:
The weird thing is that the CLI should panic in the event of a serde error, so the `schema` field in ResponseData was null (or absent). I don't remember exactly right now, but it might be `__schema` instead, I'll have time to investigate later today. If it happens with the github API as well I can probably reproduce the issue.

djc:
Here's the response from my server: https://gist.githubusercontent.com/djc/c7de96fdb8e4e6839ec3f32045408a9f/raw/e86794a82cfd13cefd831bdb9686cbd0a1ee6830/introspection-result.json

tomhoule:
Thanks! now think I know what the problem is: the root field of ResponseData should actually be `__schema` but we transform between snake and camel case, and the leading underscores most likely get swallowed. The CLI should be tested :/ I can't fix this right away, but I will do this evening - if you want to do it it's also welcome of course.

djc:
So yeah, this change seems to at least get me some output:

```diff
diff --git a/graphql_query_derive/src/codegen.rs b/graphql_query_derive/src/codegen.rs
index f9501ca..34ece7a 100644
--- a/graphql_query_derive/src/codegen.rs
+++ b/graphql_query_derive/src/codegen.rs
@@ -120,7 +121,7 @@ pub(crate) fn response_for_query(
         #variables_struct
 
         #[derive(Debug, Serialize, Deserialize)]
[...]

PR #84, opened 2018-08-22
Fix issues from #83

tomhoule:
The CLI is not at a point where it can produce useful error messages.

Credit goes to @djc for the bug report and the solution (see #83).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 rust:
   - stable
   - beta
-  - nightly
+  # - nightly
 cache: cargo
 before_script:
   - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then (rustup component add rustfmt-preview clippy-preview) fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - (breaking) Control over which types custom scalars deserialize to is given to the user: you now have to provide type aliases for the custom scalars in the scope of the struct under derive.
 - (breaking) Support for multi-operations documents. You can select a particular operation by naming the struct under derive after it. In case there is no match, we revert to the current behaviour: select the first operation.
-- Support arbitrary derives on the generated response types via the `response_derives` option on the `graphql` attribute.
+- (breaking) Support arbitrary derives on the generated response types via the `response_derives` option on the `graphql` attribute.
+
+### Fixed
+
+- Fixed codegen of fields with leading underscores - they were ignored, leading to wrong derived types for deserialization.
+- Made the CLI dump introspected schemas directly without trying to validate them.
 
 ## [0.3.0] - 2018-07-24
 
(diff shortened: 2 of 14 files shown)


djc:
With this branch, I get a number of errors that I didn't get with master just now:

```rust
warning: proc macro crates and `#[no_link]` crates have no effect without `#[macro_use]`
  --> /Users/djc/src/graphql-client/src/lib.rs:11:1
   |
11 | pub extern crate graphql_query_derive;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

   Compiling trackpay v0.1.0 (file:///Users/djc/src/trackpay)
error[E0277]: `create_entity::Currency` doesn't implement `std::fmt::Debug`
  --> tests/simple.rs:19:10
   |
19 | #[derive(GraphQLQuery)]
   |          ^^^^^^^^^^^^ `create_entity::Currency` cannot be formatted using `{:?}`
   |
   = help: the trait `std::fmt::Debug` is not implemented for `create_entity::Currency`
   = note: add `#[derive(Debug)]` or manually implement `std::fmt::Debug`
   = note: required because of the requirements on the impl of `std::fmt::Debug` for `&create_entity::Currency`
   = note: required for the cast to the object type `std::fmt::Debug`

error[E0277]: `create_entity::Currency` doesn't implement `std::fmt::Debug`
  --> tests/simple.rs:19:10
   |
19 | #[derive(GraphQLQuery)]
   |          ^^^^^^^^^^^^ `create_entity::Currency` cannot be formatted using `{:?}`
   |
   = help: the trait `std::fmt::Debug` is not implemented for `create_entity::Currency`
   = note: add `#[derive(Debug)]` or manually implement `std::fmt::Debug`
   = note: required because of the requirements on the impl of `std::fmt::Debug` for `&create_entity::Currency`
   = note: required for the cast to the object type `std::fmt::Debug`

error[E0277]: `create_entity::ResponseData` doesn't implement `std::fmt::Debug`
  --> tests/simple.rs:31:29
   |
31 |     println!("RESULT {:?}", result);
   |                             ^^^^^^ `create_entity::ResponseData` cannot be formatted using `{:?}`
   |
   = help: the trait `std::fmt::Debug` is not implemented for `create_entity::ResponseData`
   = note: add `#[derive(Debug)]` or manually implement `std::fmt::Debug`
   = note: required because of the requirements on the impl of `std::fmt::Debug` for `graphql_client::GraphQLResponse<create_entity::ResponseData>`
   = note: required by `std::fmt::Debug::fmt`

error: aborting due to 3 previous errors
```

(`Currency` is a custom enum in my project.)

djc:
Actually this now happens with both branches -- which is confusing to me, because I had this working just a few minutes earlier.

tomhoule:
It's a breaking change in the latest master: now you can place arbitrary derives on the generated response types. The only trait derived by default now is `Deserialize`, so you have to add `response_derives = "Debug"` in the `#[graphql()]` attribute on your structs.

tomhoule:
Regarding the derives, I'm not sure it's the most ergonomic choice. We could have a reasonable set of defaults and allow users to override it instead, but it feels a little bit more magic.

djc:
Alright, at least I have something working now. Thanks for your quick turnaround!

Issue #62, opened 2018-07-17
Warn when a query uses deprecated fields


tomhoule:
I can elaborate on how to implement this if somebody is interested in doing it.

tomhoule:
Idea on the interface it should offer:

`#[graphql(..., deprecated = "warn")]` (we should also support "deny)

PR #90, opened 2018-08-27
Add support for `@deprecated`.

LegNeato:
Fixes https://github.com/tomhoule/graphql-client/issues/62.


tomhoule:
Thanks for the PR! On first look it looks really nice - I think you are using the built-in rust `#[deprecated]` mechanisms which I haven't used yet but sound like the right solution - I hadn't thought of that. I'll try to find time to read it carefully as soon as possible.

LegNeato:
Yep! I am essentially translating graphql semantics to rust semantics

PR #92, opened 2018-09-02
Enable set authorization header

h-michael:
I want set authorization header, when I use graphal-client for cli.
ex)
` graphql-client introspect-schema https://api.github.com/graphql --authorization "Bearer xxxxxx"
`


tomhoule:
Thanks for the PR! It looks good, I'll wait for CI to pass and merge right away.

Issue #97, opened 2018-09-07
Improve error messages


mathstuf:
I'm seeing error messages like `operation type not in schema`. Getting the operation type name in there would be handy :) .

tomhoule:
Good point, it shouldn't be too hard to do either. We should probably have a longer, more explanatory message that says:

- operation type not in schema
- what the operation we did not find is
- what operations _are_ there
- what to do about it

I'll probably manage to find time to do this soon, if someone wants to contribute this it's even better :)

brainlessdeveloper:
Regarding this error:

```
            panic!(
                "operation type '{:?}' not in schema",
                operation.operation_type
            );
```

I can't manage to trigger that through the GitHub example - operation types are `query`, `mutation`, `subscription` or `fragment` right? The parser fails before that error is reachable. Maybe I'm misunderstanding something.

mathstuf:
IIRC, it is hit if you say `query { someUnknownField }`. It was a typo on my part, but the error message didn't help :) .

PR #100, opened 2018-09-10
codegen: improve error messages

mathstuf:
This is an initial step. Further improvements could offer suggestions
for what is available.

See #97.

### DIFF
--- a/graphql_query_derive/src/fragments.rs
+++ b/graphql_query_derive/src/fragments.rs
@@ -13,7 +13,12 @@ impl GqlFragment {
     pub(crate) fn to_rust(&self, context: &QueryContext) -> Result<TokenStream, ::failure::Error> {
         let derives = context.response_derives();
         let name_ident = Ident::new(&self.name, Span::call_site());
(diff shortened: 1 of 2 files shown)


graphql_query_derive/src/fragments.rs
```diff
@@ -13,7 +13,12 @@ impl GqlFragment {
     pub(crate) fn to_rust(&self, context: &QueryContext) -> Result<TokenStream, ::failure::Error> {
         let derives = context.response_derives();
         let name_ident = Ident::new(&self.name, Span::call_site());
-        let object = context.schema.objects.get(&self.on).expect("oh, noes");
+        let opt_object = context.schema.objects.get(&self.on);
+        let object = if let Some(object) = opt_object {
+            object
+        } else {
+            panic!("fragment '{}' cannot operate on unknown type '{}'", self.name, self.on);
+        };
```
tomhoule: That's a lot better - I forgot we still had error messages as bad as the original one :/

mathstuf: Well, it was easy to grep for to replace when I hit it :) .

PR #104, opened 2018-09-10
graphql_query_derive: support aliases in queries

mathstuf:
When querying a very abstract API (e.g., "find me database entry X"),
the name of fields may be better named in the query side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Improved some codegen error messages, giving more context. Thank @mathstuf!
 
+- Aliases in queries are now supported.
+
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


mathstuf:
I left types as the schema-preferred name since that better reflects the *type* of an alias as well anyways.

tomhoule:
Wow great! This is the oldest open issue (#5). We definitely want to support this. I went through the PR quickly, the approach seems to be the right one. I'll try to make time for a full review asap.

I am not completely sure at the moment, but there is a use case that might require to use the alias name as the struct name: one important feature with aliases is to query the same field on the same object twice with different arguments, as shown in the [official guide](https://graphql.org/learn/queries/#aliases).

tomhoule:
@h-michael This is an interesting PR :)

mathstuf:
Hrm, indeed. I'll look at getting the alias into the structure name then (a first pass failed).

mathstuf:
Type name now uses the alias as well.

mathstuf:
And now works with `cargo test --all`.

PR #105, opened 2018-09-10
WIP: tests: add a test for inline fragments

mathstuf:
---
So it seems that a node query with *just* a fragment reference is unsupported. Adding an additional field works. Here's a test which shows the difference. Not sure how to fix this at the moment.


tomhoule:
Please ping me if you need any assistance with that, I think I understand the bug.

mathstuf:
I *think* it has to do with this line:

https://github.com/tomhoule/graphql-client/blob/10f6ce7e986d9d77fc8786a6f3e136c35db6949b/graphql_query_derive/src/shared.rs#L134

but I haven't dug in too deep. Thoughts would be helpful :) .

mathstuf:
Found the culprit. I made fragment references work, but inline ones still don't. The error message now mentions that it is unimplemented.

tomhoule:
Sorry I don't have time for a full review right now (I'll do it asap). I thought inline fragments (e.g. `...on Character { name }`) only worked on uniions and interfaces. We should check the spec to see if it's valid on objects.

tests/fragments.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    let valid_inline_fragment =
+        serde_json::from_value::<inline_fragment::ResponseData>(valid_response).unwrap();
+
+    assert_eq!(
+        valid_inline_fragment.in_fragment.unwrap(),
+        "value"
+    );
+}
+*/
```
tomhoule: Looks good! Let's just delete the commented part and merge.

mathstuf:
Rebased and removed the commented test (and associated query).

PR #108, opened 2018-09-11
api: remove `GraphQL` prefix from type names

### LINKED ISSUE #101: Rename public structs
IMO, the types exposed by the crate have a bad "stutter" problem. It'd be much easier to just rename:

  - `GraphQLQuery` → `Query`
  - `GraphQLBody` → `Body`
  - `GraphQLResponse` → `Response`
  - `GraphQLError` → `Error`

collisions should be dealt with on the `use` side of things (e.g., a crate may rename them back to `GraphQL` prefixes, it may be easier to use a `Http` prefix, or possibly just use them scoped).

mathstuf:
Fixes #101.


LegNeato:
I don't think you should change the derive names as they aren't imported explicitly and  things like "Query" in large codebases are hard to grep for.

mathstuf:
@LegNeato Makes sense. Updated.

tomhoule:
There seems to be a rough consensus in favour of renaming so let's merge this :)

PR #109, opened 2018-09-11
graphql_query_derive: support deriving for input variables as well

### LINKED ISSUE #103: Variables should impl Clone
For pagination queries, it can be nice to make a "model" variables block and just update the "pagination" field where possible. Since all of the primitive types in GraphQL `impl Clone`, it should be possible to just stick `impl Clone` on these types (and probably `impl PartialEq` as well, maybe `Eq`?). Bonus points for detecting when `impl Copy` is possible as well.

mathstuf:
This just hijacks the `reponse_derives` attribute right now since it
controls derives for generated data structures and `Variables` shares
much of that.

Fixes #103.

---
Should `response_derives` be renamed to `extra_derives` since it now applies to every generated type?

### DIFF
--- a/graphql_query_derive/src/query.rs
+++ b/graphql_query_derive/src/query.rs
@@ -1,6 +1,7 @@
 use deprecation::DeprecationStrategy;
 use failure;
 use fragments::GqlFragment;
+use itertools::Itertools;
 use operations::Operation;
 use proc_macro2::Span;
 use proc_macro2::TokenStream;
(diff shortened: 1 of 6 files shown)


graphql_query_derive/src/query.rs
```diff
@@ -71,6 +75,12 @@ impl QueryContext {
             ));
         }
 
+        self.variables_derives.extend(
+            attribute_value
+                .split(',')
+                .map(|s| s.trim())
+                .map(|s| Ident::new(s, Span::call_site())),
+        );
```
tomhoule: So we derive the same traits as for the response at the moment, that seems like a good start. I think we might want to enable users to choose different derives for variables and ResponseData (since the use cases are different) in the future, but it's a good start. Great!

tomhoule:
I'll happily merge after the branch is rebased :)

tomhoule:
I'll also document it in the changelog.

mathstuf:
Rebased and changelog entry added.

PR #111, opened 2018-09-11
Cache input files

### LINKED ISSUE #110: Cache file contents
Instead of reading the schema and query every time, it'd be better to cache the contents of files (and ideally their parsed out contents) to help with compile times.

mathstuf:
---
Fixes #110.

### DIFF
--- a/graphql_query_derive/Cargo.toml
+++ b/graphql_query_derive/Cargo.toml
@@ -11,6 +11,7 @@ proc-macro = true
 
 [dependencies]
 failure = "0.1"
+lazy_static = "1.0"
 quote = "^0.6"
 syn = "0.14"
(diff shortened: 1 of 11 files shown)


mathstuf:
Timing this topic on a crate which imports a single query and schema file 16 times via `cargo check --all` after a compile and then `touch src/queries.rs`:

  - Before 60s
  - After 58s

So maybe not that much performance improvement. Anyone else have projects which can test this out?

mathstuf:
I suspect that #114 would help with compile times more. I'm up to 20+ queries and 2 minute `cargo check` times now.

tomhoule:
Looks good, merged, thanks a lot!

PR #112, opened 2018-09-12
graphql_query_derive: improve "no such field" error message

mathstuf:
See #97.


graphql_query_derive/src/shared.rs
```diff
@@ -110,8 +112,15 @@ pub(crate) fn response_fields_for_selection(
...
                     .iter()
-                    .find(|field| field.name.as_str() == name.as_str())
-                    .ok_or_else(|| format_err!("Could not find field: {}", name.as_str()))?;
+                    .find(|field| &field.name == name)
+                    .ok_or_else(|| {
+                        format_err!(
+                            "Could not find field `{}` on `{}`. Available fields: `{}`.",
+                            name.as_str(),
+                            type_name,
+                            schema_fields.iter().map(|ref field| &field.name).format("`, `"),
```
tomhoule: the `format` method is really neat!

mathstuf: Yeah, it makes things like this really easy.

PR #99, opened 2018-09-09
Add new crate, graphql_client_codegen

h-michael:
Implementation of #43.

- Cut out the main logic of graphql_query_derive into graphql_client_codegen for sharing logic with graphql_client_cli.

@tomhoule 
Is this as you intended?

PR #116, opened 2018-09-14
Only output required structs

### LINKED ISSUE #114: [RFE] Only output required types in generated module
Some kind of marking which schema types are actually used when generating the other types could be useful.

mathstuf:
---
`cargo check` times goes from 110–120 seconds → 7 seconds for my crate :) .

Fixes #114.


tomhoule:
That's a huge improvement! I'll review the PR asap (tomorrow at the latest)

tomhoule:
We just merged #99 - it moves most of the code from the graphql-query-derive crate to a new graphql-client codegen crate so we can explore other ways to do code generation (through the cli, build scripts or even as a library compiled to wasm that can be published as an npm package :) ). As a side-effect most of the files in graphql_query_derive have moved but they haven't changed.

The conflicts message from github is scary but I tried to rebase this branch locally and since it's just file renaming (mostly) there are only two lines with a merge conflict apparently. I'll start reviewing now.

PR #119, opened 2018-09-16
Specify graphql_client_codegen version

h-michael:
- specify dependency version

### DIFF
--- a/graphql_query_derive/Cargo.toml
+++ b/graphql_query_derive/Cargo.toml
@@ -12,4 +12,4 @@ proc-macro = true
 [dependencies]
 syn = "0.14"
 proc-macro2 = { version = "0.4", features = [] }
-graphql_client_codegen = { path = "../graphql_client_codegen/" }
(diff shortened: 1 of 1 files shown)


tomhoule:
It makes sense indeed. I don't have enough experience with publishing multi-crate projects, but I think we should push graphql_client_codegen to crates.io at the same time as the next release of graphql-client (hopefully in the next 10 days, I'm going to give a small presentation to a local meetup about the library and it would be nice to have a new release - there are a lot of awesome unreleased features now). Do you have more experience/ideas on that topic?

h-michael:
@tomhoule 
Now we have the same version of each crate,
but what if we make breaking changes only on dependent crate?

tomhoule:
I'll try to read up on this - but I think it's maybe good to keep all crates at the same version because users should only import `graphql_client` and not the others.

h-michael:
> I'll try to read up on this

Sorry, what's this you said ?

tomhoule:
Maybe there are blog posts or articles with good solutions to this problem.

Issue #43, opened 2018-06-29
CLI command for codegen


tomhoule:
It would be useful for people who want to generate the code in advance to get better editor support (RLS for example does not take into account generated code from derives yet).

Linking `proc_macro2` could be a challenge, from my past experience.

h-michael:
@tomhoule 
I want this feature for code completion.
I think implementing code generation feature is possible by graphql_query_derive.
Do you have any ideas?

tomhoule:
So in the current state, all the code generation logic lives in `graphql_query_derive` indeed. The idea would be to first refactor the top level functions in `lib.rs`, `schema.rs` and `query.rs` so we can expose a function like this:

```rust
pub fn generate_module(
  query: graphql_parser::query::Document,
  schema: graphql_query_derive::Schema,
  options: GraphQLClientDeriveOptions,
) -> TokenStream
```

The query is parsed directly from the `.graphql` file, the schema is parsed either from a `.graphql` or a `.json` file (there are `From` impls in `schema.rs`), and the options are currently extracted from the `#[graphql()]` attribute but they would have to be taken as CLI arguments.

The name `GraphQLClientDeriveOptions` is something I just came up with, we could find something better.

There are two caveats:

- I don't remember if it is possible to have other exports than the derive macros in a macro crate. If that is the case we would need to split codegen out of `graphql-query-derive` into a new `graphql-client-codegen` crate. This is not hard but it is work.

- Since we are working with `TokenStream`, I don't know if we can avoid dependency on the `proc-macro` crate, which is part of the compiler and cannot be statically linked. If we can't, this means the codegen part of the CLI will only work if you have the Rust compiler installed locally (and I am afraid of versioning problems there). This is an acceptable drawback however because presumably, if you are using `graphql-client`, you have a Rust compiler.

This is a refactoring we will need to do at some point because it enables other features (like using the codegen part of the library as a build script, if people are interested).

h-michael:
@tomhoule 
> I don't remember if it is possible to have other exports than the derive macros in a macro crate. If that is the case we would need to split codegen out of graphql-query-derive into a new graphql-client-codegen crate. This is not hard but it is work.

We cannot other exports that has not `proc_macro_derive`.
So, I split codegen out of graphql-query-derive into a new graphql-client-codegen crate, first of all.

tomhoule:
It sounds like the right thing to do. I just reserved the crate name on crates.io so you can use it.

[...]

PR #122, opened 2018-09-17
Enable generating graphql client code by cli

h-michael:
Implement #43

- [x] implement feature

### DIFF
--- a/graphql_client_cli/src/main.rs
+++ b/graphql_client_cli/src/main.rs
@@ -4,12 +4,15 @@ extern crate reqwest;
 extern crate structopt;
 #[macro_use]
 extern crate graphql_client;
+extern crate graphql_client_codegen;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde;
 extern crate serde_json;
 
 use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE, ACCEPT};
+use std::fs::File;
+use std::io::Write as IoWrite;
 use std::path::PathBuf;
 use structopt::StructOpt;
 
@@ -38,9 +41,21 @@ enum Cli {
     #[structopt(name = "generate")]
     Generate {
         // should be a glob
-        paths: String,
+        /// Path to graphql query file
         #[structopt(parse(from_os_str))]
-        schema: PathBuf,
+        query_path: PathBuf,
+        /// Path to graphql schema file
+        #[structopt(parse(from_os_str))]
+        schema_path: PathBuf,
+        /// Name of struct
+        selected_operation: String,
+        /// Additional derives
+        /// --additional-derives='Serialize,PartialEq'
+        #[structopt(short = "a", long = "additional-derives")]
+        additional_derives: Option<String>,
+        /// allow, deny, or warn
+        #[structopt(short = "d", long = "deprecation-strategy",)]
+        deprecation_strategy: Option<String>,
         #[structopt(parse(from_os_str))]
         output: PathBuf,
     },
@@ -55,10 +70,20 @@ fn main() -> Result<(), failure::Error> {
             authorization,
         } => introspect_schema(schema_location, output, authorization),
         Cli::Generate {
-            paths: _,
-            schema: _,
-            output: _,
-        } => unimplemented!(),
+            query_path,
+            schema_path,
+            selected_operation,
+            additional_derives,
+            deprecation_strategy,
+            output,
+        } => generate_code(
+            query_path,
+            schema_path,
+            selected_operation,
+            additional_derives,
+            deprecation_strategy,
+            output,
+        ),
     }
 }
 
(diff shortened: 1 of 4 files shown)


h-michael:
@tomhoule 
How do you think this implementation?
I am glad if you give me your opinion.

graphql_client_cli/src/main.rs
```diff
@@ -39,9 +43,15 @@ enum Cli {
...
         #[structopt(parse(from_os_str))]
-        schema: PathBuf,
+        query_path: PathBuf,
+        #[structopt(parse(from_os_str))]
+        schema_path: PathBuf,
+        selected_operation: String,
+        #[structopt(short = "a", long = "additional_derives")]
+        additional_derives: Option<String>,
+        #[structopt(short = "d", long = "deprecation_strategy",)]
+        deprecation_strategy: Option<u8>,
```
tomhoule: For additional derives and deprecation strategy, I prefer hyphens instead of underscores for the `long` names, so we can write `graphql-client-cli generate --additional-derives=Serialize,Clone` instead of `--additional_derives`.

For the deprecation strategy options could we take a string instead? I think `--deprecation-strategy=allow` is more intuitive than `--deprecation-strategy=0`.

graphql_client_codegen/src/lib.rs
```diff
@@ -55,8 +55,10 @@ lazy_static! {
         CacheMap::default();
 }
 
-pub struct GraphQLClientDeriveOptions<'a> {
-    pub input: &'a syn::DeriveInput,
+pub struct GraphQLClientDeriveOptions {
+    pub selected_operation: String,
+    pub additional_derives: Option<String>,
+    pub deprecation_strategy: Option<deprecation::DeprecationStrategy>,
```
tomhoule: Since there is default (`Warn`) I think `deprecation_strategy` does not need to be an option, what do you think?

graphql_client_cli/src/main.rs
```diff
@@ -39,9 +43,15 @@ enum Cli {
     #[structopt(name = "generate")]
     Generate {
         // should be a glob
-        paths: String,
         #[structopt(parse(from_os_str))]
-        schema: PathBuf,
+        query_path: PathBuf,
+        #[structopt(parse(from_os_str))]
+        schema_path: PathBuf,
+        selected_operation: String,
```
tomhoule: I think we could make selected_operation optional (in the `derive` implementation we just take the first one by default if no operation is selected).

graphql_client_codegen/src/lib.rs
LegNeato: I actually think there is a "bug". The default on the struct is "allow" but in usage is "warn". I'll make them the same in a PR in the morning.

h-michael:
> Have you tried this on your projects?
Not yet, but generated test deriven file. I'll try it after done my business : )

graphql_client_cli/src/main.rs
h-michael: OK I'll fix it later :)

tomhoule: There's no rush :)

graphql_client_codegen/src/lib.rs
h-michael: > I think deprecation_strategy does not need to be an option, what do you think?

Now, we determine default value in the method, [generate_module_token_stream](https://github.com/tomhoule/graphql-client/pull/122/files/0b00534b6e037cb1dc8ebeed38dfb886745e5e98#diff-4cc81fc19b88d044ccd920473bc5492fR81).
So I let this field option.

graphql_client_cli/src/main.rs
h-michael: @tomhoule 
I think selected_option is also used as module name.
So we couldn't let selected_option be optional.

h-michael:
@tomhoule 
I fixed the parts you pointed out and reply to comments.

h-michael:
I've generated rust code as sample.

```rust
pub mod test_strcut {
    #![allow(non_camel_case_types)]
    #![allow(non_snake_case)]
    #![allow(dead_code)]
    use serde;
    pub const QUERY: &'static str = "query loginUser {\n  viewer {\n    login\n  }\n}\n\n";
    pub const OPERATION_NAME: &'static str = "TestStrcut";
    #[allow(dead_code)]
    type Boolean = bool;
    #[allow(dead_code)]
    type Float = f64;
    #[allow(dead_code)]
    type Int = i64;
    #[allow(dead_code)]
    type ID = String;
    #[derive(Deserialize)]
    #[doc = "A user is an individual\'s account on GitHub that owns repositories and can make new content."]
    pub struct RustLoginUserViewer {
        #[doc = "The username used to login."]
        pub login: String,
    }
    #[derive(Serialize)]
    pub struct Variables;
    #[derive(Deserialize)]
    pub struct ResponseData {
        #[doc = "The currently authenticated user."]
        pub viewer: RustLoginUserViewer,
    }
}
impl<'de> ::graphql_client::GraphQLQuery<'de> for TestStrcut {
    type Variables = test_strcut::Variables;
    type ResponseData = test_strcut::ResponseData;
    fn build_query(
        variables: Self::Variables,
    ) -> ::graphql_client::GraphQLQueryBody<Self::Variables> {
        ::graphql_client::GraphQLQueryBody {
            variables,
            query: test_strcut::QUERY,
            operation_name: test_strcut::OPERATION_NAME,
        }
    }
}
```

h-michael:
BTW, I'm not good at English, so I want to someone or you help me make cli option's help messages.

tomhoule:
That looks really good :) I'll try to find the time so we can do the CLI help messages asap (probably this evening for me so in 10 hours or so).

graphql_client_cli/src/main.rs
```diff
@@ -38,9 +41,21 @@ enum Cli {
...
         // should be a glob
-        paths: String,
+        /// Path to graphql query file
         #[structopt(parse(from_os_str))]
-        schema: PathBuf,
+        query_path: PathBuf,
+        /// Path to graphql schema file
+        #[structopt(parse(from_os_str))]
+        schema_path: PathBuf,
+        /// Name of struct
```
tomhoule: I have ideas on how we might change this in the future (because we can basically generate code for all operations, so I just opened issue #134. It will be a big refactoring and it needs to be discussed so let us merge this PR first :)

graphql_client_cli/src/main.rs
```diff
@@ -38,9 +41,21 @@ enum Cli {
...
+        selected_operation: String,
+        /// Additional derives
```
tomhoule: nitpick: we could add a bit more info: "Additional derives that will be added to the generated structs and enums for the response and the variables"

graphql_client_cli/src/main.rs
```diff
@@ -38,9 +41,21 @@ enum Cli {
...
+        /// --additional-derives='Serialize,PartialEq'
+        #[structopt(short = "a", long = "additional-derives")]
+        additional_derives: Option<String>,
+        /// allow, deny, or warn
```
tomhoule: nitpick: we could add something like "default: warn"

h-michael:
@tomhoule 
Thank you for giving me advice. 
I fixed help messages and I've understood the issue(#134).

h-michael:
After this PR is merged, I'll make a PR that update clid README.md.

tomhoule:
Writing this here so one of us does it: we should also change the sentence about the CLI in the top-level README, because it is useful now.

h-michael:
@tomhoule OK. Should I write down it before merge PR ?

tomhoule:
No I think we can merge as soon as CI passes. It's not urgent.

PR #124, opened 2018-09-17
Upgrade graphql-parser to v0.2.2

h-michael:
There is no breaking change.
[v0.2.0...v0.2.2](https://github.com/graphql-rust/graphql-parser/compare/v0.2.0...v0.2.2)


mathstuf:
Just curious, was the [default of `^`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) used by `cargo` not sufficient to get a newer one in a local `Cargo.lock`?

tomhoule:
Indeed, I guess in this case we could have kept a lower required version.

Issue #121, opened 2018-09-17
Upgrade reqwest above v0.8.x


h-michael:
We need upgrade reqwest to v0.9.0-pre, because of https://github.com/seanmonstar/reqwest/issues/345 .
But this version is not stable.
When we upgrade reqwest?
Now users who use openssl-1.1.1 can not build.

tomhoule:
Fixed by #132

PR #132, opened 2018-09-18
Upgrade reqwest to ^0.9.0

h-michael:
related #121

reqwest v0.8.x don't correspond to openssl 1.1.1.
So we should upgrade reqwest.
There are [some breaking changes](https://github.com/seanmonstar/reqwest/blob/master/CHANGELOG.md#breaking-changes).

refer to: https://docs.rs/reqwest/0.9.0/reqwest/struct.RequestBuilder.html#method.headers

### DIFF
--- a/graphql_client_cli/src/main.rs
+++ b/graphql_client_cli/src/main.rs
@@ -9,8 +9,7 @@ extern crate serde_derive;
 extern crate serde;
 extern crate serde_json;
 
-use reqwest::header::*;
-use reqwest::mime;
+use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE, ACCEPT};
 use std::path::PathBuf;
 use structopt::StructOpt;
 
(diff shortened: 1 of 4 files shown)


graphql_client_cli/src/main.rs
```diff
@@ -101,13 +100,8 @@ fn introspect_schema(
...
-    let mut headers = Headers::new();
-    headers.set(Accept(vec![qitem(mime::APPLICATION_JSON)]));
-
-    match authorization {
-        Some(token) => headers.set(reqwest::header::Authorization(token)),
-        None => (),
-    };
+fn construct_headers() -> HeaderMap {
+    let mut headers = HeaderMap::new();
+    headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
```
tomhoule: in addition to Content-Type, we should also set the [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) header (to `application/json` as well).

h-michael: @tomhoule Oops! Thank you for finding my oversight.

h-michael:
@tomhoule I fixed it.

tomhoule:
I'll merge as soon as CI passes

PR #136, opened 2018-09-21
Resolove warning in test code

h-michael:
Resolving warning in test code.
But I don't know how to resolve deprecated warning.
So I resolve other than that.

### DIFF
--- a/tests/custom_scalars.rs
+++ b/tests/custom_scalars.rs
@@ -16,6 +16,7 @@ type NetworkAddress = Ipv4Addr;
     query_path = "tests/custom_scalars/query.graphql",
     schema_path = "tests/custom_scalars/schema.graphql"
 )]
+#[allow(dead_code)]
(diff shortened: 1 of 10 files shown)


mathstuf:
Note that I took `dead_code` out in #116 (prior to that the warnings were quite formidable). I think it'd be better to add some mention of the top-level structs in the code itself to resolve those warnings.

h-michael:
@mathstuf Thanks your comments.
I push new commit. How about this?

mathstuf:
Thanks, that looks better. This also allows the test to be indicators for "hey, you're outputting types that aren't necessary".

h-michael:
@mathstuf
BTW, do you have any idea how to remove deprecated warning in test?

mathstuf:
There's an RFC (can't find it at the moment) which mentions `#[expect(warning_name)]`, but other than that, no.

PR #150, opened 2018-09-26
Fix wrong operationName when no operation is selected (fixes #149)

### LINKED ISSUE #149: We may be sending incorrect `operationName`s
See discusion in #147

tomhoule:
For example if we have this:

query SelectName {
  person { name }
}

query SelectAge {
  person { age }
}

And our code looks like this:

  query_path = "..",
  schema_path = "...",
)]
struct SomeQuery;

then the operation we generate code for will default to the first one, and operationName should be SelectName.


tomhoule:
@h-michael If you are interested a review is very welcome :) This PR made me realize we should improve the control flow in the codegen code, at the top level (before we have the query and schema and generate code for them), it's a bit confused.

h-michael:
@tomhoule OK wait a little for review 👍

tomhoule:
There's no hurry

graphql_client_codegen/src/codegen.rs
```diff
@@ -8,10 +8,29 @@ use query::QueryContext;
...
+            _ => (),
+        }
+    }
+
+    operations
+        .iter()
+        .find(|op| op.name == struct_name)
+        .map(|i| i.to_owned())
+        .or_else(|| operations.iter().next().map(|i| i.to_owned()))
+}
```
tomhoule: I'll document that before merging, so people don't need to read the implementation to know what it does.

graphql_client_codegen/src/lib.rs
```diff
@@ -93,6 +93,16 @@ pub fn generate_module_token_stream(
...
 
+    // Determine which operation we are generating code for. This will be used in operationName.
+
+    let operation = if let Some(op) = codegen::select_operation(&query, &options.struct_name) {
+        op
+    } else {
+        panic!("Query document defines no operation.")
+    };
+
+    let operation_string = operation.name.clone();
```
tomhoule: that can be just `&operation.name`

graphql_client_codegen/src/operations.rs
```diff
@@ -113,3 +113,41 @@ impl ::std::convert::From<OperationDefinition> for Operation {
...
+                    .variable_definitions
+                    .iter()
+                    .map(|v| v.clone().into())
+                    .collect(),
+                selection: (&s.selection_set).into(),
+            },
+            OperationDefinition::SelectionSet(_) => panic!(SELECTION_SET_AT_ROOT),
+        }
+    }
+}
```
tomhoule: this is mostly copy-pasted, ideally I'll find time to clean it up before merging

graphql_query_derive/src/lib.rs
```diff
@@ -37,7 +37,7 @@ fn build_graphql_client_derive_options(input: &syn::DeriveInput) -> GraphQLClien
         .unwrap_or(deprecation::DeprecationStrategy::default());
 
     GraphQLClientDeriveOptions {
-        selected_operation: input.ident.to_string(),
+        struct_name: input.ident.to_string(),
```
tomhoule: I renamed selected_operation to struct_name because we are probably going to generate code for _all_ the operations in a query document in codegen other than the derive.

tests/operation_selection.rs
```diff
@@ -38,11 +49,45 @@ fn operation_selection_works() {
...
+    let expected_heights = heights::ResponseData {
+        mountain_height: Some(224),
+        building_height: Some(12),
+    };
 
-    let expected_echo = r##"ResponseData { echo: Some("tiramisù") }"##;
-    let expected_heights =
-        r##"ResponseData { mountain_height: Some(224), building_height: Some(12) }"##;
+    assert_eq!(expected_echo, echo_response_data);
+    assert_eq!(expected_heights, heights_response_data);
```
tomhoule: I should add an assertion: `assert_eq!(expected_heights_unrelated, heights_response_data);` (to check that response shapes are the same.

h-michael:
I think that it is better for users to be able to select operation_name regardless of struct_name, like this.

```
query SelectName {
  person { name }
}

query SelectAge {
  person { age }
}
```
```
#[derive(GraphQLQuery)]
#[graphql(
  query_path = "..",
  schema_path = "...",
  operation_name = "SelectAge"
)]
struct Person;
```

If users not define operation_name or query file includes only one query, we select first operation name.

This is just idea.

```
Is a user define operation name? ▶️ yes ▶️ We select this as a operation name.
⬇️
No
⬇️ 
Does target query file include only one query? ▶️ yes ▶️ We select this as a operation name.
⬇️
No
⬇️
Does target query file include query that name is same as struct name? ▶️ yes ▶️ We select this as a operation name.
⬇️
No
⬇️
We select first query name
```

h-michael:
@tomhoule I've just done review.

tomhoule:
Thanks for the review! I think the behaviour you describe is the one we have right now - I'm checking with the CLI to see if anything is broken.

tomhoule:
I am running into difficulties with linking libproc_macro when building the CLI. There are issues to explore solutions - I think we may be able to find a solution.

```
../target/release/graphql-client: error while loading shared libraries: libproc_macro-d0651654f325a058.so: cannot open shared object file: No such file or directory
```

h-michael:
@tomhoule 

I also faced the same problem today.
That is probably a problem with the rust compiler.
FYI https://github.com/alexcrichton/proc-macro2/issues/115

tomhoule:
Apparently we may just have to wait, this is very good news.

tomhoule:
For now the CLI should work exactly the same as before this PR, right?

h-michael:
@tomhoule 
> For now the CLI should work exactly the same as before this PR, right?

I think it works same behavior.
Is there anything you do not think about in this PR that I think is the following?

1. Providing method to implement multiple queries in one struct.
2. Providing method to make struct name and operation name different names.

tomhoule:
For point 2 we could definitely provide an option in the `graphql` attribute, like you said. We should maybe create an issue to see how it interacts with struct names (I like the current behaviour, so I'd like to keep it as a fallback).

For point 1 I don't know how that would look with the derive, but I have ideas about how to make this nice with the CLI and build script codegen flows, I'll try to write them down soon.

h-michael:
> I like the current behaviour, so I'd like to keep it as a fallback

I think so too.

> For point 1 I don't know how that would look with the derive, but I have ideas about how to make this nice with the CLI and build script codegen flows, I'll try to write them down soon.

OK 👍

PR #159, opened 2018-10-02
Enable test over target wasm32-unknown-unknown for nightly builds

### LINKED ISSUE #120: Check that the library works with the wasm32-unknown-unknown target in CI
I think `cargo test --all` tests all targets, so we would only need to `rustup target add wasm32-unknown-unknown` beforehand. Otherwise we can `cargo test --all --target=wasm32-unknown-unknown`.

maxymshg:
This PR aims to resolve #120 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ rust:
   - stable
   - beta
   - nightly
+matrix:
+  allow_failures:
+    - rust: nightly
 cache: cargo
 before_script:
   - if [ "$TRAVIS_RUST_VERSION" = "stable" ]; then (rustup component add rustfmt-preview clippy-preview) fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,11 @@ rust:
 cache: cargo
 before_script:
   - if [ "$TRAVIS_RUST_VERSION" = "stable" ]; then (rustup component add rustfmt-preview clippy-preview) fi
+  - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then (rustup target add wasm32-unknown-unknown) fi
 script:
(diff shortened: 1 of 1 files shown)


tomhoule:
The nightly builds are failing because of a dependency of `prettytable` apparently. Since I think we only use it for examples it may work with `cargo test` (without the `--all`).

maxymshg:
I suggest to use matrix feature and allow failure on nightly builds. What do you think?

tomhoule:
The problem with doing it this way is that we won't catch regressions in nightly anymore. I'd prefer to test selectively (not all crates) for that target.

maxymshg:
I looked over different options, and cannot find how to exclude/include specific targets for modules. You mentioned that you would prefer to test selectively for that target, could you please share for which modules we should do such testing?

LegNeato:
Might be worthwhile to do what juniper does and use cargo-make.

maxymshg:
Thank you, @LegNeato . I'll take a look, I never used `cargo-make` before.

tomhoule:
At work now but I'll try to investigate too asap :)

tomhoule:
Sorry I haven't had time to work on this in the last days. So I don't think we should allow builds to fail on nightly (unless nightly is temporarily broken for us, it happens). I don't know the scope of what we can test with that target, so the 2 options are: run the tests we can explicitly, maybe from a script, or use cargo-make, which is the same but more powerful.

I am not sure I want to introduce the additional complexity unless it's necessary however. Let us try to see if we can test the `graphql-client` crate only with wasm32-unknown-unknown first. I may give it a try this weekend.

tomhoule:
I am not sure this is needed anymore, since we have tests for graphql_client_web.

PR #164, opened 2018-10-06
Remove the lifetime parameter from GraphQLQuery

tomhoule:
It makes it easier to work with async clients, since the 'de lifetime is scoped to the deserializer.

I would be interested in having feedback on this - I don't know if I might be breaking something. In my current effort on the browser client this makes things easier.


tomhoule:
By the way, this is a breaking change.

PR #168, opened 2018-10-07
Make it possible to generate formatted code

h-michael:
I make it possible to generate formatted code.

When users install cli with option like this,
`cargo install graphql_client_cli --features rustfmt`,
graphql-client generate command generate formatted code by rustfmt.

If users don't want to generate formatted code, they should execute command with `--no-formatting`

### DIFF
--- a/graphql_client_cli/src/generate.rs
+++ b/graphql_client_cli/src/generate.rs
@@ -10,6 +10,7 @@ pub fn generate_code(
     selected_operation: String,
     additional_derives: Option<String>,
     deprecation_strategy: Option<String>,
+    no_formatting: bool,
     output: PathBuf,
 ) -> Result<(), failure::Error> {
     let deprecation_strategy = deprecation_strategy.as_ref().map(|s| s.as_str());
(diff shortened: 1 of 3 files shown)


h-michael:
I confused whether to make it features or not.

tomhoule:
Sorry for the late review! I think we should make it an optional feature, because rustfmt adds quite a bit of dependencies and people may want to use custom formatting.

graphql_client_cli/src/main.rs
```diff
@@ -46,8 +49,13 @@ enum Cli {
         additional_derives: Option<String>,
         /// You can choose deprecation strategy from allow, deny, or warn.
         /// Default value is warn.
-        #[structopt(short = "d", long = "deprecation-strategy",)]
+        #[structopt(short = "d", long = "deprecation-strategy")]
         deprecation_strategy: Option<String>,
+        /// If you don't want to execute rustfmt to generated code, set this option.
+        /// Default value is false.
+        /// Formmating feature is disabled as default installation.
```
tomhoule: nitpick: typo on "Formatting"

graphql_client_cli/src/generate.rs
```diff
@@ -25,8 +26,44 @@ pub fn generate_code(
...
+        let mut out = Vec::with_capacity(codes.len() * 2);
+
+        Session::new(config, Some(&mut out))
+            .format(Input::Text(codes.to_string()))
+            .unwrap_or_else(|err| panic!("rustfmt error: {}", err));
+
+        return String::from_utf8(out).unwrap();
+    }
+    #[cfg(not(feature = "rustfmt"))]
+    codes.to_string()
```
tomhoule: Since `format` should be called when the feature is disabled, we could even replace this with `unreachable!()`.

graphql_client_cli/src/generate.rs
```diff
@@ -25,8 +26,44 @@ pub fn generate_code(
...
+
+    if cfg!(feature = "rustfmt") && !no_formatting {
+        let codes = format(&codes);
+        write!(file, "{}", codes);
+        return Ok(());
+    }
+
+    write!(file, "{}", codes);
+
     Ok(())
```
tomhoule: It's not super important, but the `write!` and `Ok(())` lines are repeated. We could write `let code = if cfg!(...) && !no_formatting { .... } else { .... };` and have the `write` and `Ok` only once.

h-michael:
@tomhoule 
Thanks for your review, and I have fixed and rebased it.

tomhoule:
Thanks!

PR #169, opened 2018-10-08
Update example github schema.graphql

h-michael:
Update github schema.graphql.


mathstuf:
It might be worth committing a script to do this so that it is done in a consistent way. [Here](https://gitlab.kitware.com/utils/rust-ghostflow/blob/master/ghostflow-github/src/graphql/update_schema.sh) is the script I wrote to do it for my project (which includes preview `Accept` headers for preview APIs).

PR #174, opened 2018-10-17
Add a graphql_client_web crate for browser usage

tomhoule:
What needs to be done before this is mergeable:

- [x] Merge the PR to remove the lifetime parameter from graphql-query
- [x] Document it (much) better
- [x] Refine the errors, provide better causes
- [x] Make the `ClientError` enum non-exhaustive
- [x] Make sure it does not panic
- [x] Review the dependencies
- [x] Enable setting custom http headers - and have better defaults there
- [ ] Crucially, test this in CI (we are going to have to run a browser there)
- [ ] Write a better wasm example


tomhoule:
@h-michael  As you can see there is a lot of work left, but if you want to review that would be welcome :)

h-michael:
@tomhoule Thanks for mentioning to me!
I'm going to review.

graphql_client_web/tests/web.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[macro_use]
+extern crate serde_derive;
+
+use futures::Future;
+use graphql_client_web::Client;
+use wasm_bindgen::JsValue;
+use wasm_bindgen_test::*;
+
+use std::panic;
```
tomhoule: can be removed

graphql_client_web/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+/// - use it to perform queries with the [call] method
+pub struct Client {
+    endpoint: String,
+}
+
+/// All the ways a request can go wrong.
+///
+/// not exhaustive
+#[derive(Debug, Fail)]
+#[non_exhaustive]
```
h-michael: You should add `#![feature(non_exhaustive)]`, otherwise I can't build.

Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [workspace]
 members = [
+    "graphql_client_cli",
+    "graphql_client_codegen",
     "graphql_client",
     "graphql_client/examples/example_module",
     "graphql_client/examples/github",
     "graphql_query_derive",
```
h-michael: missing "graphql_client_web"

graphql_client_web/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn client_new() {
+        Client::new("https://example.com/graphql");
+        Client::new("/graphql");
```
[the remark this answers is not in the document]
tomhoule: It's just to check that it doesn't crash

graphql_client_web/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "graphql_client_web"
+version = "0.1.0"
```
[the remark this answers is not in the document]
tomhoule: Good point, I'll change

graphql_client_web/src/lib.rs
tomhoule: thanks, done

theduke:
Out of curiosity, why will this be a separate crate rather than a normal feature?

tomhoule:
It only works on a specific target (wasm32-unknown-unknown) so it looked like it would be complicated to configure. It also has a very specific setup for tests so it seemed to make sense to make it a separate crate.

If the configuration is straightforward we could reexport graphql-client-web as a feature on graphql-client easily, so I don't really see any big drawback to making it a separate crate. I'm open to other ideas of course :)

tomhoule:
Also as a more general comment, we want client crates for many different platforms (probably graphql-client-node, graphql-client-async, graphql-client-sync, ...) so I didn't want to clutter the core crate too much.

tomhoule:
I just realized reexporting from graphql_client wouldn't work because this crate depends on the types from graphql_client. That's a usability problem.

theduke:
There is some similiar discussion going on here: https://github.com/seanmonstar/reqwest/issues/288

theduke:
Just wanted to say: I tried out this branch on a wasm project, and it works like a charm. :fireworks:

Issue #134, opened 2018-09-20
CLI codegen ergonomics


tomhoule:
At the moment the CLI generates one module for one file. The general use-case however is probably projects with more than one query, and having to issue more than one command seems a bit too complicated.

The most use similar CLI tool at the moment is [apollo-cli](https://github.com/apollographql/apollo-cli) and I think we could take a similar approach.

Invoking graphql-client-cli could look something like this:

`graphql-client-cli generate --schema-path=./schema.json --deprecations=warn 'src/**/*.graphql'`, and it would generate one file per .graphql file, with one module per query.

The shared code could live in one module for the whole project - we can achieve a lot more reuse with this approach than with the derive-based interface.

Let's discuss what interface we want for the CLI in this isue.

tomhoule:
@h-michael tagging you since you implemented the CLI codegen - in case you want to discuss this.

h-michael:
I agree with this idea.
But only one thing to consider is how to handle the `--additional - derives` option.
Although it may be a trivial problem, some people may want to divide what they want to derive.
If the policy is decided, I would like to start implementing samples that divide multiple modules into one file.

tomhoule:
There is one library with a similar problem (although it's a build script): [prost](https://github.com/danburkert/prost).

[Relevant doc page](https://docs.rs/prost-build/0.3.2/prost_build/struct.Config.html) (in particular the `field_attribute` and `type_attribute` functions. I don't find the solution very elegant and would like to find something better if we can, but it works.

tomhoule:
One other solution we could explore is giving options like `additional_derives` as annotation or directives inside the `.graphql` files.

h-michael:
> One other solution we could explore is giving options like additional_derives as annotation or directives inside the .graphql files.

Even if it is an annotation, I do not think it is a good thing to describe what is not in the specification.
If we do not provide it with the option of cli, is not it better to provide a config file for graphql-client-cli?

For example, how about this?

```bash
graphql-client-cli generate --schema-path=./schema.json --deprecations=warn 'src/**/*.graphql' \
  --common-additional-derives='Debug,Clone' \
  --additional-derives='FooTrait:TargetStruct1,TargetStruct2'
```

The format of the option needs to be discussed, but in essence it separates the options of deriving in common and deriving only for a specific struct.

However, as I think that it is complicated as cli, I think that it may be possible to provide a config of toml format that can handle the same format.
[...]

PR #175, opened 2018-10-19
Generate all queries at onece

h-michael:
The purpose of this PR is to allow codegen to make all queries at once.

related with #134

And implementing selected_operation option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ The default is `warn`.
 ## Query documents with multiple operations
 
 You can write multiple operations in one query document (one `.graphql` file). You can then select one by naming the struct you `#[derive(GraphQLQuery)]` on with the same name as one of the operations. This is neat, as it allows sharing fragments between operations.
+
 If you want to name the struct different from query name, you can use ``selected_operation`` argument like this.
 
 #[derive(GraphQLQuery)]
--- a/README.md
+++ b/README.md
@@ -127,6 +127,15 @@ The default is `warn`.
 ## Query documents with multiple operations
 
 You can write multiple operations in one query document (one `.graphql` file). You can then select one by naming the struct you `#[derive(GraphQLQuery)]` on with the same name as one of the operations. This is neat, as it allows sharing fragments between operations.
+If you want to name the struct different from query name, you can use ``selected_operation`` argument like this.
+
+#[derive(GraphQLQuery)]
+#[graphql(
+    schema_path = "src/graphql/schema.json",
+    query_path = "src/graphql/queries/my_query.graphql",
+    selected_operation = "SearchQuery"
+)]
+pub struct MyQuery;
 
 There is an example [in the tests](./tests/operation_selection).
 
--- a/graphql_client_cli/README.md
+++ b/graphql_client_cli/README.md
@@ -1,3 +1,61 @@
 # GraphQL client CLI
 
 This is still a WIP, the main use for it now is to download the `schema.json` from a GraphQL endpoint, which you can also do with [apollo-codegen](https://github.com/apollographql/apollo-cli).
+
+## Install
+
+```
+cargo install graphql-client-cli
+```
+
+## introsect schema
+
+```
+USAGE:
+    graphql-client introspect-schema [OPTIONS] <schema_location>
+
+FLAGS:
+    -h, --help       Prints help information
+    -V, --version    Prints version information
+
+OPTIONS:
+        --authorization <authorization>
+        --output <output>                  Where to write the JSON for the introspected schema.
+
+ARGS:
+    <schema_location>    The URL of a GraphQL endpoint to introspect.
+```
+
+## generate client code
+
+```
+USAGE:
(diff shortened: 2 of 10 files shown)


graphql_client_cli/src/main.rs
```diff
@@ -15,24 +15,24 @@ extern crate serde_json;
...
+    //     /// The URL of a GraphQL endpoint to introspect.
+    //     schema_location: String,
+    //     /// Where to write the JSON for the introspected schema.
+    //     #[structopt(parse(from_os_str))]
+    //     #[structopt(long = "output")]
+    //     output: Option<PathBuf>,
+    //     // Set authorizaiton header.
+    //     #[structopt(long = "authorization")]
+    //     authorization: Option<String>,
+    // },
```
h-michael: I make these lines to be comment out for build.
Finally, I'll rollback these.

graphql_client_codegen/src/lib.rs
```diff
@@ -144,19 +151,57 @@ pub fn generate_module_token_stream(
...
+    let mut query_struct_token_streams = vec![];
+    for (schema_output, operation_name) in schema_and_operations {
+        let token_stream =
+            build_query_struct_token_stream(module_name, schema_output, operation_name);
+        query_struct_token_streams.push(token_stream.to_string());
+    }
+    quote!(
+        #common_token_stream
+        #(#query_struct_token_streams)*
+    )
```
h-michael: @tomhoule 
It did not work as it would be string.
I really want to use [concat](https://manishearth.github.io/rust-internals-docs/syntax/tokenstream/struct.TokenStream.html), but we can only use it nightly.

tomhoule:
Thanks for the PR! I am not sure we want to generate code for all queries in the derive case however, what would the API look like?

h-michael:
> I am not sure we want to generate code for all queries in the derive case

OK, I'll fix for derive case not to generating all queries.

h-michael:
@tomhoule 
Do you think which is better derive interface that generate all queries or generate only first one query if struct name is not included queries as query name.

How about new attribute, selected-operation, for derive ?

If struct name is not included queries as query name, and do not specify selected-operation attribute,  we generate all the queries.

h-michael:
@tomhoule 
Now, ready to review.
If you think this implementation objective, I'll rebase and clean commit log.

graphql_client_codegen/src/codegen.rs
```diff
@@ -140,10 +149,20 @@ pub fn response_for_query(
...
     let response_derives = context.response_derives();
 
+    let respons_data_struct_name = if multiple_operation {
+        Ident::new(
+            format!("{}ResponseData", operation.name).as_str(),
+            Span::call_site(),
+        )
+    } else {
+        Ident::new("ResponseData", Span::call_site())
+    };
```
h-michael: If all the queries are in same module, we should give unique name to ResponseData and Variables.

tomhoule:
Sorry I haven't reviewed yet - I will as soon as I have some free time :)

README.md
```diff
@@ -127,6 +127,15 @@ The default is `warn`.
 ## Query documents with multiple operations
 
 You can write multiple operations in one query document (one `.graphql` file). You can then select one by naming the struct you `#[derive(GraphQLQuery)]` on with the same name as one of the operations. This is neat, as it allows sharing fragments between operations.
+If you want to name the struct different from query name, you can use ``selected_operation`` argument like this.
```
tomhoule: small thing: this won't show up as a separate paragraph, there needs to be a blank line between the two lines

graphql_client_cli/README.md
```diff
@@ -1,3 +1,61 @@
 # GraphQL client CLI
 
 This is still a WIP, the main use for it now is to download the `schema.json` from a GraphQL endpoint, which you can also do with [apollo-codegen](https://github.com/apollographql/apollo-cli).
+
+## Install
+
+```
+cargo install graphql-client-cli
```
tomhoule: What do you think of adding the `--force` flag? (in general users always want the latest version imo)

graphql_client_cli/README.md
```diff
@@ -1,3 +1,61 @@
...
+```
+
+## introsect schema
```
tomhoule: typo `introspect`

graphql_client_cli/README.md
```diff
@@ -1,3 +1,61 @@
...
+ARGS:
+    <query_path>     Path to graphql query file.
+    <schema_path>    Path to graphql schema file.
+    <module_name>    Name of module.
+    <output>         Path you want to output to.
+```
+
+If you want to use formatting feature, you should install like this.
+
+`cargo install graphql-client-cli --features rustfmt`
```
tomhoule: Nice idea to put the usage docs in the README!

h-michael:
@tomhoule Thanks! I've fixed it.
Should I rebase commits ?

tomhoule:
No need to, there's no conflict. Thanks a lot!

PR #211, opened 2019-01-06
Support recursive fragments

tomhoule:


tomhoule:
Update on this: there are a few subtleties in the question of whether a fragment needs to be boxed. The current solution errs on the side of too much boxing, but it can be refined later. I will apply that to inputs as well for correctness (the current solution does not handle all cases).

PR #213, opened 2019-01-12
Unpin version of graphql-parser

upsuper:
The version was pinned in #163 for avoid using new version of `combine`. But this somehow makes downstream users not able to use a newer version of `combine`.

We can see if this is stil a problem.


upsuper:
It seems to work just fine. Is there any reason we still need to pin the version?

h-michael:
@tomhoule 
This is difference.
https://github.com/graphql-rust/graphql-parser/compare/89a260c0dbf792c1d9c849710cefba2f401a6c6e...13346fe2226dbfee5547ca7b5fba37b4a0343518

And I think @upsuper said is this.
https://github.com/graphql-rust/graphql-parser/compare/89a260c0dbf792c1d9c849710cefba2f401a6c6e...13346fe2226dbfee5547ca7b5fba37b4a0343518
#diff-80398c5faae3c069e4e6aa2ed11b28c0R17

I think we don't have any reasons that we freeze version.

tomhoule:
Thanks for checking! I don't remember exactly but I think it was because at some point, a newer version of combine didn't compile on nightly, or something like that. If CI passes I don't see any problem with changing this. Let's merge this, thank you!

PR #214, opened 2019-01-12
Use default root operation type names when not specified

upsuper:
According to [the GraphQL spec](https://facebook.github.io/graphql/June2018/#sec-Root-Operation-Types):
> While any type can be the root operation type for a GraphQL operation, the type system definition language can omit the schema definition when the `query`, `mutation`, and `subscription` root types are named `Query`, `Mutation`, and `Subscription` respectively.
> 
> Likewise, when representing a GraphQL schema using the type system definition language, a schema definition should be omitted if it only uses the default root operation type names.

Actually, GitHub's introspection nowadays doesn't include schema definition at all.

### DIFF
--- a/graphql_client_codegen/src/codegen.rs
+++ b/graphql_client_codegen/src/codegen.rs
@@ -69,20 +69,19 @@ pub(crate) fn response_for_query(
     }
 
     let response_data_fields = {
-        let opt_root_name = operation.root_name(&context.schema);
-        let root_name: &str = if let Some(root_name) = opt_root_name {
-            root_name
+        let root_name = operation.root_name(&context.schema);
+        let opt_definition = context
+            .schema
+            .objects
+            .get(&root_name);
+        let definition = if let Some(definition) = opt_definition {
+            definition
         } else {
             panic!(
                 "operation type '{:?}' not in schema",
                 operation.operation_type
             );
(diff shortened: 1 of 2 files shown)


tomhoule:
Looks good, I hadn't seen that part of the spec, indeed.

The only doubt I have is the following: with this PR, we would look for a schema definition, and fall back to the default names. But could there be cases where there is no schema definition _and_ those are not defined? Probably not if the schema data comes from introspection (it would be completely invalid). I also don't think the error message would be too bad in that case (probably something about not finding the Query type).

I am also thinking about how this could be tested. I can probably do that later if you need this change asap.

upsuper:
So, if we fall back to the default name, but the corresponding type doesn't exist, it would just panic with something like `operation type 'Query' not in schema`, which should be clear enough.

upsuper:
I can workaround this... so no hurry. I was thinking about testing this as well, but did't come up with something trivial, so I decided to just do the code work for now :)

upsuper:
Ping for this PR. Is there anything you are waiting for?

tomhoule:
Nope, I just haven't had time in the past days. Thanks for the ping, let's merge this!

upsuper:
Thanks!

PR #216, opened 2019-01-12
Always derive PartialEq and Eq for enums

upsuper:
In GraphQL, enums can only have atom values, so we can always derive `PartialEq` and `Eq` for them. Although this can probably be done via specifying `response_derives`, but it may not always be desired to have the whole response `PartialEq` and `Eq`.

### DIFF
--- a/graphql_client_codegen/src/query.rs
+++ b/graphql_client_codegen/src/query.rs
@@ -6,7 +6,7 @@ use proc_macro2::Span;
 use proc_macro2::TokenStream;
 use schema::Schema;
 use selection::Selection;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use syn::Ident;
 
 /// This holds all the information we need during the code generation phase.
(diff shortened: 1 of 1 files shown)


graphql_client_codegen/src/query.rs
```diff
@@ -120,21 +120,21 @@ impl<'query, 'schema> QueryContext<'query, 'schema> {
     }
 
     pub(crate) fn response_enum_derives(&self) -> TokenStream {
-        let enum_derives: Vec<_> = self
+        let always_derives = [
+            Ident::new("Eq", Span::call_site()),
+            Ident::new("PartialEq", Span::call_site()),
+        ];
+        let mut enum_derives: BTreeSet<_> = self
```
upsuper: `BTreeSet` is used here so that we don't have duplicate derives, also it would order the derives when iterated.

upsuper:
Ping for this PR.

tomhoule:
Seems reasonable, let's merge this. Thanks for the ping.

PR #222, opened 2019-02-07
Fix path to graphql_client in GitHub example

rhysd:
I confirmed my fix make the example work again on my local machine (macOS 10.12.6)


tomhoule:
It's broken indeed, thanks for the PR!

PR #229, opened 2019-03-13
Make generated code less confusing.

tomhoule:
Lengthier pull request message incoming, but there are two headlines:

- In the derive workflow, the struct name _must_ match an operation name, and
  this is the operation that will be selected. This makes naming in generated
  code less confusing.
- In the CLI workflow, we generate one file per query document (.graphql file),
  with one module per operation.


graphql_client_codegen/src/lib.rs
```diff
@@ -91,8 +88,21 @@ pub fn generate_module_token_stream(
...
+    let operations = match (operations, &options.mode) {
+        (Some(ops), _) => ops,
+        (None, &CodegenMode::Cli) => codegen::all_operations(&query),
+        (None, &CodegenMode::Derive) => {
+            return Err(
+                format_err!("The struct name does not match any defined operation in the query file.\nStruct name: {}\nDefined operations: {}", options.struct_ident().map(|i| i.to_string()).as_ref().map(String::as_str).unwrap_or(""), query.definitions.iter().filter_map(|definition| match definition { graphql_parser::query::Definition::Operation(op) => match op { graphql_parser::query::OperationDefinition::Mutation(m) => Some(m.name.as_ref().unwrap()),
+graphql_parser::query::OperationDefinition::Query(m) => Some(m.name.as_ref().unwrap()),
+graphql_parser::query::OperationDefinition::Subscription(m) => Some(m.name.as_ref().unwrap()),
+graphql_parser::query::OperationDefinition::SelectionSet(_) => panic!("Bare selection sets are not supported"),
+        }, _ => None }).fold(String::new(), |mut acc, item| { acc.push_str(&item); acc.push_str(", "); acc }).trim_end_matches(", ")),
```
tomhoule: This is very ugly, it needs to be extracted into a function.

tomhoule:
@h-michael @theduke I would be interested in your feedback, especially on the CLI part.

The idea now is that if you generate code for the following document:

```graphql
# In my_query.graphql

query QueryA { .... }
query QueryB { .... }
mutation SomeMutation { ... }
```

The CLI generated code will look like this:

```rust
// In my_query.rs

pub struct QueryA;

pub mod query_a { ... }

pub struct QueryB;

pub mod query_b { ... }

pub struct SomeMutation;

pub mod some_mutation { ... }
```

That way the operations don't live in the same module, there is no risk of names clashing anymore.

h-michael:
@tomhoule I'll see tonight :)

graphql_client_codegen/src/generated_module.rs
```diff
@@ -0,0 +1,86 @@
...
+                    const __QUERY_WORKAROUND: &str = include_str!(#path);
+                )
+            })
+            .unwrap_or_else(|| quote! {});
+
+        let query_string = &self.query_string;
+        let impls = self.build_impls()?;
+
+        let struct_declaration = match self.options.mode {
+            CodegenMode::Cli => quote!(#module_visibility struct #operation_name_ident),
```
h-michael: need semicolon.
```suggestion
            CodegenMode::Cli => quote!(#module_visibility struct #operation_name_ident;),
```

graphql_client_cli/src/main.rs
```diff
@@ -69,8 +68,6 @@ enum Cli {
         /// Default value is pub.
         #[structopt(short = "m", long = "module_visibility")]
         module_visibility: Option<String>,
-        #[structopt(parse(from_os_str))]
-        output: PathBuf,
```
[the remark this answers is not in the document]
tomhoule: Good point, it's a valid thing to want, I'll reimplement it as output directory.

graphql_client_codegen/src/generated_module.rs
tomhoule: Indeed, thanks!

tomhoule:
For the operation selection from the CLI, I agree we can keep it. Do you think it should be the same layout as "all operations" but restricted to the selected operation, or something different?

h-michael:
I think that default behavior is generate operation is better.
How about adding `--specify-modules` option?

tomhoule:
I am not sure how `--specify-modules` would work, can you elaborate?

tomhoule:
Merging and releasing this now, as the proposed changes would be additive.

PR #231, opened 2019-03-27
Fix naming clash in the CLI flags (fixes #230)

### LINKED ISSUE #230: CLI uses the same shortcut twice
```
$ graphql-client generate --help
OPTIONS:
    -o, --output-directory <output_directory>
            The directory in which the code will be generated.
    -o, --selected-operation <selected_operation>
            Name of target query. If you don't set this parameter, cli generate all queries in query file.
```
Two different options are using `-o` as a shortcut in 0.7, confused me a little.

tomhoule:


tomhoule:
Hmm, let's merge this then. There must have been a breaking change in clap/structopt.

PR #232, opened 2019-03-29
Remove #[cfg(tests)] from build.rs

h-michael:
See https://github.com/budziq/rust-skeptic/issues/60

I tried to use https://github.com/assert-rs/docmatic.
But we can't use that, because we use monorepo.

### DIFF
--- a/graphql_client/build.rs
+++ b/graphql_client/build.rs
@@ -1,9 +1,6 @@
 extern crate skeptic;
 
 fn main() {
-    #[cfg(tests)]
-    {
-        // Generates doc tests for the readme.
-        skeptic::generate_doc_tests(&["../README.md"]);
-    }
+    // Generates doc tests for the readme.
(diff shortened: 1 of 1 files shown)


tomhoule:
The CI failure is due to rustfmt.

I don't know how much longer skeptic makes the builds, but I think the priority is fixing this so we can merge this PR once CI passes.

Thanks a lot for working on this!

h-michael:
@tomhoule I've fixed it.

> I don't know how much longer skeptic makes the builds, but I think the priority is fixing this so we can merge this PR once CI passes.

I don't care of build time. Many people dislike depending of skeptic at production build.

tomhoule:
Yes it's a good point. It would be nice to get rid of the dependency, but having the readme tested is very useful. Maybe we can do something with rustdoc directly, we should investigate.

PR #233, opened 2019-03-31
codegen: improve "could not find field" error message

brainlessdeveloper:
I've considered adding newlines in between but I don't know if there's any existing convention that would conflict with that.

### DIFF
--- a/graphql_client_codegen/src/shared.rs
+++ b/graphql_client_codegen/src/shared.rs
@@ -6,6 +6,19 @@ use proc_macro2::{Ident, Span, TokenStream};
 use query::QueryContext;
 use selection::*;
 
+fn format_available_fields(fields: &[GqlObjectField]) -> String {
+    fields
+        .iter()
+        .map(|ref field| &field.name)
+        .fold(String::new(), |mut acc, item| {
+            acc.push_str(item);
+            acc.push_str(", ");
+            acc
+        })
+        .trim_end_matches(", ")
+        .to_owned()
+}
+
 pub(crate) fn render_object_field(
     field_name: &str,
     field_type: &TokenStream,
@@ -77,7 +90,13 @@ pub(crate) fn field_impls_for_selection(
                 let ty = fields
                     .iter()
                     .find(|f| &f.name == name)
-                    .ok_or_else(|| format_err!("could not find field `{}`", name))?
+                    .ok_or_else(|| {
+                        format_err!(
+                            "Could not find field `{}`. Available fields: `{}`.",
+                            name,
+                            format_available_fields(fields).as_str()
+                        )
+                    })?
                     .type_
                     .inner_name_str();
                 let prefix = format!("{}{}", prefix.to_camel_case(), alias.to_camel_case());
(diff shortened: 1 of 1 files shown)


brainlessdeveloper:
```
error: proc-macro derive panicked
  --> src/main.rs:23:10
   |
23 | #[derive(GraphQLQuery)]
   |          ^^^^^^^^^^^^
   |
   = help: message: Code generation failed.
           Could not find field `bananas`. Available fields: `__typename, codeOfConduct, codesOfConduct, license, licenses, marketplaceCategories, marketplaceCategory, marketplaceListing, marketplaceListings, meta, node, nodes, organization, rateLimit, relay, repository, repositoryOwner, resource, search, topic, user, viewer`.    
```

graphql_client_codegen/src/shared.rs
```diff
@@ -6,6 +6,19 @@ use proc_macro2::{Ident, Span, TokenStream};
 use query::QueryContext;
 use selection::*;
 
+fn format_available_fields(fields: &[GqlObjectField]) -> String {
+    fields
+        .iter()
+        .map(|ref field| &field.name)
+        .fold(String::new(), |mut acc, item| {
+            acc.push_str(item);
+            acc.push_str(", ");
```
mathstuf: I think `Itertools::format` would just be better.

```rust
fields.iter().map(|ref field| &field.name).format("`, `")
```

This keeps the field names enclosed in "`", but leaves the commas without markup.

brainlessdeveloper:
That does look better. Added as an additional commit so it can be reviewed separately.

graphql_client_codegen/src/shared.rs
```diff
@@ -77,7 +82,13 @@ pub(crate) fn field_impls_for_selection(
                 let ty = fields
                     .iter()
                     .find(|f| &f.name == name)
-                    .ok_or_else(|| format_err!("could not find field `{}`", name))?
+                    .ok_or_else(|| {
+                        format_err!(
+                            "Could not find field `{}`. Available fields: `{}`.",
+                            name,
+                            format_available_fields(fields).as_str()
```
mathstuf: This `as_str` seems superfluous now (same below).

brainlessdeveloper: Right - thanks!

tomhoule:
Looks good! We discussed the iter formatting issue on another channel - we should address that concern, rebase and then let's merge this.

brainlessdeveloper:
So I should rebase and then do the thing without `itertools`?

tomhoule:
yep - for the `itertools` part, you can implement a `Display` helper, something like:

```rust
struct CommaSeparated<T>(T);

impl<A, B> std::fmt::Display for CommaSeparated<A>
where A: Iter<Item=B>,
  B: Display {
 ....
}
```

tomhoule:
If it's too hard we can think about keeping the itertools dependency.

mathstuf:
Is this something we should revisit? Improving error messages is almost always a good thing IMO.

tomhoule:
> Is this something we should revisit? Improving error messages is almost always a good thing IMO.

yeah definitely. We could have an issue so we don't forget

PR #249, opened 2019-08-15
Derive Debug for ResponseData and Variables

### LINKED ISSUE #248: Variables should impl Debug
I think `dbg!(&vars)` is very useful for debugging purposes.
Perhaps it should be introduced as variables_derives param on `graphql` procmacro?

MikailBag:
fixes #248

### DIFF
--- a/graphql_client_codegen/src/inputs.rs
+++ b/graphql_client_codegen/src/inputs.rs
@@ -220,7 +220,7 @@ mod tests {
         };
 
         let expected: String = vec![
-            "# [ derive ( Clone , Serialize ) ] ",
+            "# [ derive ( Clone , Debug , Serialize ) ] ",
             "pub struct Cat { ",
             "pub offsprings : Vec < Cat > , ",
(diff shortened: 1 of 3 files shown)


tomhoule:
Thanks for the PR!

Adding Debug to variables is fine as a temporary workaround, but we already have the `response_derives` ResponseData and we don't want Debug by default there because it has a non-negligible effect on code size when targeting browsers. 

With that change we could merge this.

I'll try to implement the proper solution (`variable_derives`) for the next release.

mathstuf:
Hmm. Looking at the code, `additional_derives` applies to variables and responses. Any reason not to just use that? Or should that be split to have `variable` and `response` specific halves?

tomhoule:
@mathstuf  I haven't touched that code in such a long time that I don't remember what the current state is. I agree that variables and response derives should be separate - I'll try to make it happen for next release (soon™)

tomhoule:
This should be addressed by https://github.com/graphql-rust/graphql-client/pull/285, which was just merged. Please open an issue if doesn't work for your use-case :)

PR #253, opened 2019-08-21
Update all dependencies, especially syn and rustfmt to 1.0+

### LINKED ISSUE #246: Update to syn, quote and proc-macro2 1.0
The 1.0 versions of these crates just came out, we should upgrade.

davidgraeff:
Fixes #246

I have updated all dependencies via `cargo upgrade --all` first (using the cargo-edit crate).
`graphql_query_derive/src/attributes.rs` required some intention because one method was renamed and one struct field type changed.

`cargo build` and `cargo test` works.

Signed-off-by: David Graeff <david.graeff@web.de>

### DIFF
--- a/graphql_client_codegen/src/codegen_options.rs
+++ b/graphql_client_codegen/src/codegen_options.rs
@@ -13,7 +13,6 @@ pub enum CodegenMode {
 }
 
 /// Used to configure code generation.
-#[derive(Debug)]
 pub struct GraphQLClientCodegenOptions {
     /// Which context is this code generation effort taking place.
     pub mode: CodegenMode,
--- a/examples/github/Cargo.toml
+++ b/examples/github/Cargo.toml
@@ -7,16 +7,16 @@ edition = "2018"
 [dependencies]
 failure = "*"
 graphql_client = { path = "../../graphql_client" }
-serde = "1.0"
-serde_derive = "1.0"
-serde_json = "1.0"
-reqwest = "^0.9.0"
-prettytable-rs = "0.7.0"
-structopt = "0.2.10"
-dotenv = "0.13.0"
-envy = "0.3.2"
-log = "0.4.3"
-env_logger = "0.5.10"
+serde = "^1.0"
+serde_derive = "^1.0"
+serde_json = "^1.0"
+reqwest = "^0.9"
+prettytable-rs = "^0.7"
+structopt = "^0.2"
+dotenv = "^0.13"
+envy = "^0.3"
+log = "^0.4"
+env_logger = "^0.5"
 
 [workspace]
 members = ["."]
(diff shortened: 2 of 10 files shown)


tomhoule:
Thanks for the PR!

The stable build is failing on rustfmt - that should be straightforward to fix.

I was thinking about whether pinning everything to the latest version like this was going to cause more duplicate dependencies for users of the library (the new version requirements are much stricter), but I don't think it's a big problem. I haven't found any good resources on best practices for libraries. So let's merge this as soon as the build is green :)

davidgraeff:
I have formatted the code with `rustfmt`. The rust vscode extension fomatting that I had used is probably considering my vscode preferences and not the project preferences.

And you are right, for most libraries the patch version should not be specified. I have changed that to major.minor now and added minor-version-semver compatibility via the caret character, eg *failure = "^0.1^*.

I kept *rustfmt-nightly* with major.minor.patch, because that library had a breaking change on one of my own projects after upgrading from "1.4.x" to "1.4.5".

tomhoule:
Sounds good!

[GitHub is having an outage today](https://www.githubstatus.com/), let's wait this out before merging :)

davidgraeff:
I have now included a few more bits that I require for my own project.

Rust keyword escaping happens for (almost) all parts of a schema now, an integration test is part of this PR. It's still just around 200 lines of changes, so maybe you could review this in one go? :)
Splitting into two PRs would be feasible as well though, as you like.

h-michael:
@davidgraeff Thank you for your PR :)

I think this PR includes some different purpose changes.

1. update dependencies
2. fix compile error for +1.37
3. fix codegen

If you divide PR for each issue, I think that it will be merged faster as a result.

davidgraeff:
Yeah I guess so. Wanted it to be conserved here first and for some first opinions. Fortunately I have split the PR in feature commits.

graphql_client_codegen/src/codegen_options.rs
```diff
@@ -28,6 +30,7 @@ pub struct GraphQLClientCodegenOptions {
     /// The deprecation strategy to adopt.
     deprecation_strategy: Option<DeprecationStrategy>,
     /// Target module visibility.
+    #[derivative(Debug = "ignore")]
```
davidgraeff: The new crate is used to allow the "Debug" derive on this struct by ignoring this field.
Because syn's Visibility no longer derives of Debug, the code doesn't compile otherwise.

You might as well remove the Debug trait from this struct

davidgraeff:
See the inline comment :)

graphql_client_codegen/src/codegen_options.rs
tomhoule: If we can live without the debug, let's do that. Otherwise a wrapper struct with a dummy Debug implementation would work. Sorry to bother you with that but I feel like it's worth it to keep the dependency count low.

davidgraeff: Alright I'll just remove the Debug trait for now.

PR #257, opened 2019-08-26
Adding the normalize_query_types feature to normalize all generated to CamelCase.

### LINKED ISSUE #254: Scalar types that begin with a lowercase letter
I am having issues with a scalar type in the graphql schema that beings with a lowercase letter as the type is not valid in rust. Is there any way to work around this?

### LINKED ISSUE #255: Code gen: Redefinitions
I'm having issues with the code generation part. See the following example:

`common.graphql` Query file:
```graphql
query ping_query {ping}
```
`schema.graphql` Schema file:
```graphql
type Query {
  """Returns the literal string 'pong'."""
  ping: String!
}
```

Generated Rust code:
```rust
pub struct ping_query;
pub mod ping_query {
    #![allow(dead_code)]
    pub const OPERATION_NAME: &'static str = "ping_query";
    pub const QUERY: &'static str = "query ping_query {ping}";
    use serde::{Deserialize, Serialize};
    #[allow(dead_code)]
    type Boolean = bool;
    #[allow(dead_code)]
    type Float = f64;
    #[allow(dead_code)]
    type Int = i64;
    #[allow(dead_code)]
    type ID = String;
    #[derive(PartialEq, Serialize)]
    pub struct Variables;
    #[derive(Deserialize, PartialEq, Serialize)]
    pub struct ResponseData {
        #[doc = "Returns the literal string 'pong'.\n"]
        pub ping: String,
    }
}
impl graphql_client::GraphQLQuery for ping_query {
    type Variables = ping_query::Variables;
    type ResponseData = ping_query::ResponseData;
    fn build_query(variables: Self::Variables) -> ::graphql_client::QueryBody<Self::Variables> {
        graphql_client::QueryBody {
            variables,
            query: ping_query::QUERY,
            operation_name: ping_query::OPERATION_NAME,
        }
    }
}
```

I don't know about older compiler versions, but 1.37+ fail to compile this with:
```
error[E0428]: the name `ping_query` is defined multiple times
 --> src/queries/common.rs:2:1
  |
1 | pub struct ping_query;
  | ---------------------- previous definition of the type `ping_query` here
2 | pub mod ping_query {
  | ^^^^^^^^^^^^^^^^^^ `ping_query` redefined here
  |
  = note: `ping_query` must be defined only once in the type namespace of this module

error[E0428]: the name `ping_query` is defined multiple times
 --> src/queries/common.rs:2:1
  |
1 | pub struct ping_query;
  | ---------------------- previous definition of the type `ping_query` here
2 | pub mod ping_query {
  | ^^^^^^^^^^^^^^^^^^ `ping_query` redefined here
```

Apparently the module name and struct name are named the same.

And I found out why:
```rust
let module_name = Ident::new(&self.operation.name.to_snake_case(), Span::call_site());
let operation_name_ident = Ident::new(&self.operation.name, Span::call_site());
```
If the query is already in snake case, the above mentioned error happens.
Instead operation_name should be always converted to camel case like so:

```rust
let module_name = Ident::new(&self.operation.name.to_snake_case(), Span::call_site());
let operation_name_ident =
            Ident::new(&self.operation.name.to_camel_case(), Span::call_site());
```

Fixed in #253

markcatley:
This probably isn't the final solution to this issue but I made this change so that I could continue with what I was implementing. At a minimum, the readme would need to be updated to describe the feature.

I have added the normalize_query_types that normalises types (structs, type aliases, etc) generated by the code-gen to use CamelCase as it is my understanding that this is idiomatic rust. I have done this as a feature as it's potentially a breaking change.

I was having issues with scalars as described in #254 but then noticed that other types were non-idiomatic. 

Fixes #254 and fixes #255.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,5 @@ script:
   - cargo test --all
   - cargo build --manifest-path=./examples/github/Cargo.toml
   - cargo build --manifest-path=./examples/web/Cargo.toml
+  - cargo build --manifest-path=./examples/hasura/Cargo.toml
   - if [ "$TRAVIS_RUST_VERSION" = "stable" ]; then (xvfb-run cargo test --manifest-path=./graphql_client/Cargo.toml --features="web" --target wasm32-unknown-unknown) fi
--- a/examples/hasura/Cargo.toml
+++ b/examples/hasura/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "graphql_query_hasura_example"
+version = "0.1.0"
+authors = ["Mark Catley <mark@catley.net.nz>"]
+edition = "2018"
+
+[dependencies]
+failure = "*"
+graphql_client = { path = "../../graphql_client", features = [ "normalize_query_types" ] }
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+reqwest = "^0.9.0"
+prettytable-rs = "0.7.0"
+structopt = "0.2.10"
+dotenv = "0.13.0"
+envy = "0.3.2"
+log = "0.4.3"
+env_logger = "0.5.10"
+
+[workspace]
+members = ["."]
--- a/examples/hasura/README.md
+++ b/examples/hasura/README.md
@@ -1,0 +1,3 @@
+# graphql-client Hasura examples
+
+The schema was generated using (Hasura)[https://hasura.io/]. It is here to demonstrate the normalize_query_types feature and would require some work to create a Hasura instance that matches the schema. It is primarily present to ensure the feature compiles correctly.
--- a/examples/hasura/src/main.rs
+++ b/examples/hasura/src/main.rs
@@ -1,0 +1,67 @@
+// use failure::*;
+use graphql_client::*;
+use log::*;
+use prettytable::*;
+// use structopt::StructOpt;
+
+type Bpchar = String;
+type Timestamptz = String;
+
+#[derive(GraphQLQuery)]
+#[graphql(
+    schema_path = "src/schema.graphql",
+    query_path = "src/query_1.graphql",
+    response_derives = "Debug"
+)]
+struct UpsertIssue;
+
+fn main() -> Result<(), failure::Error> {
+    use upsert_issue::{IssuesUpdateColumn::*, *};
+    dotenv::dotenv().ok();
+    env_logger::init();
+
+    let q = UpsertIssue::build_query(Variables {
+        issues: vec![IssuesInsertInput {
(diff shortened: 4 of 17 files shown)


davidgraeff:
Seems like you solve #255 indeed in a more comprehensive way than I did. I'm not sure if it should be hidden behind a feature flag. It actually breaks generated code and the transformations should always happen imo. I don't think backwards compatibility is an issue because this could never have worked for anyone before.

Why do you introduce an online test and don't follow the _cli tests structure with schema+query files?

markcatley:
Thank you.

I am definitely happy for it not to be behind a feature flag, however, it is definitely a breaking change. For example, if you go into the github example and turn the feature flag on then you get the following error:
```
error[E0412]: cannot find type `Uri` in module `super`
  --> src/main.rs:10:10
   |
10 | #[derive(GraphQLQuery)]
   |          ^^^^^^^^^^^^ not found in `super`

For more information about this error, try `rustc --explain E0412`.
error: Could not compile `graphql_query_github_example`.
```

I really wasn't sure how to test this so I created the new example so it can most likely be done in a better way. I don't follow where you're suggesting it should be tested. What is the path to the files you're talking about?

davidgraeff:
Tests are in `graphql_client/tests`. I would have expected most of them to be in the _codegen directory like the github test, but I'm not really opinionated.

tomhoule:
Sorry for the delayed reply, I'm on vacations at the moment :)

I also think this should be the default. It's unambiguous that we should convert from snake case to camel case. The one in the GitHub example is more puzzling, as it's a renaming from `URI` to `Uri`, which is not intuitive. We could think about only doing the case transformation if

- there is an underscore in the type  (excluding the introspection types maybe), or
- it starts with a lowercase letter

What do you think?

tomhoule:
We should also document it (I could do it if you don't have the time).

davidgraeff:
> there is an underscore in the type (excluding the introspection types maybe), or
it starts with a lowercase letter

If you introduce rules for the transformation, there will be people tripping over those one or two odd named types in their schemas. I'd say the transformation should happen all the time, imo.

tomhoule:
Sorry for the delay! I think we should merge this without the feature, I'll try to find the time to rebase and merge (it's great if you want to rebase yourself of course :))

tomhoule:
Update: I'm fixing the merge conflicts right now.

tomhoule:
Merged in 26db903178bde6b26511e3c7278685a6e95f864b and added to the changelog :) We'll probably make this the default eventually. Thanks a lot!

mathstuf:
See #283 for some problems I saw when trying to address #281. PR incoming shortly.

PR #258, opened 2019-08-27
Fix structopt usage in the CLI tool

davidgraeff:
Depends on #253

Fix structopt usage in the CLI tool. Potential breaking change!
    
structopt uses the first letter for a "short" option.
"selected_operation" (with "o") as well as "output_directory" (with "out")
claimed "-o" which resulted in:
    
      thread 'main' panicked at 'Argument short must be unique
    	-o is already in use
    
Fixed by removing the short option for "selected_operation" and keep "-o" for output directory.

This fixes the fix introduced in 0.7.1. I guess that `structopt` got more strict after the version update.

### DIFF
--- a/examples/github/Cargo.toml
+++ b/examples/github/Cargo.toml
@@ -7,16 +7,16 @@ edition = "2018"
 [dependencies]
 failure = "*"
 graphql_client = { path = "../../graphql_client" }
-serde = "1.0"
-serde_derive = "1.0"
-serde_json = "1.0"
-reqwest = "^0.9.0"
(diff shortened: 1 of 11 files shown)


tomhoule:
Merged in 1509faaab9e60844db981649298db545f5126add - thanks!

davidgraeff:
Sweet. I can rebase the other PRs later the day if you like me to.

tomhoule:
Sorry for the late reply. We can see case by case, I'll try to find time to re-review and see if there's anything else to address in the coming days. But thanks :)

PR #259, opened 2019-08-27
Catch more cases where a rust keyword in schemas or queries would break code generation

### LINKED ISSUE #255: Code gen: Redefinitions

### LINKED ISSUE #256: Rust keywords are not escaped in all situations
Currently code generation produces invalid rust code for a schema like this:

```graphql
schema {
  query: Query
  mutation: Mutation
}

"""This directive allows results to be deferred during execution"""
directive @defer on FIELD

"""The top-level Query type."""
type Query {
  """Keyword type"""
  search: Self
}

"""Keyword type"""
type Self {
  """
  A keyword variable name with a keyword-named input type
  """
  transactions(struct: extern!): Result
}

"""Keyword type"""
type Result {
  """Keyword field."""
  for: String
  """dummy field with enum"""
  status: AnEnum
}


"""Keyword input"""
input extern {
  """A field"""
  id: crate
}

"""Input fields for searching for specific values."""
input crate {
  """Keyword field."""
  enum: String

  """Keyword field."""
  in: [String!]
}


"""Enum with keywords"""
enum AnEnum {
  where
  self
}
```

davidgraeff:
Depends on #253
Depends on #258
Fixes #255
Fixes #256

Catch more cases where a rust keyword in schemas or queries would break code generation. This includes variables, Inputs, enum fields, response fields. Adds integration test (keywords_query.graphql+keywords_schema.graphql).

This PR makes the following schema possible (contains rust keywords and doesn't work right now):

```graphql
"""The top-level Query type."""
type Query {
  """Keyword type"""
  search: Self
}

"""Keyword type"""
type Self {
  """
  A keyword variable name with a keyword-named input type
  """
  transactions(struct: extern!): Result
}

"""Keyword type"""
type Result {
  """Keyword field."""
  for: String
  """dummy field with enum"""
  status: AnEnum
}


"""Keyword input"""
input extern {
  """A field"""
  id: crate
}

"""Input fields for searching for specific values."""
input crate {
  """Keyword field."""
  enum: String

  """Keyword field."""
  in: [String!]
}


"""Enum with keywords"""
enum AnEnum {
  where
  self
}
```

### DIFF
--- a/examples/github/Cargo.toml
+++ b/examples/github/Cargo.toml
@@ -7,16 +7,16 @@ edition = "2018"
 [dependencies]
 failure = "*"
 graphql_client = { path = "../../graphql_client" }
-serde = "1.0"
-serde_derive = "1.0"
-serde_json = "1.0"
-reqwest = "^0.9.0"
-prettytable-rs = "0.7.0"
-structopt = "0.2.10"
-dotenv = "0.13.0"
-envy = "0.3.2"
-log = "0.4.3"
-env_logger = "0.5.10"
+serde = "^1.0"
+serde_derive = "^1.0"
+serde_json = "^1.0"
+reqwest = "^0.9"
+prettytable-rs = "^0.7"
+structopt = "^0.2"
+dotenv = "^0.13"
+envy = "^0.3"
+log = "^0.4"
+env_logger = "^0.5"
 
 [workspace]
 members = ["."]
--- a/graphql_client/Cargo.toml
+++ b/graphql_client/Cargo.toml
@@ -10,26 +10,26 @@ categories = ["network-programming", "web-programming", "wasm"]
 edition = "2018"
 
 [dependencies]
-failure = "0.1"
+failure = "^0.1"
 graphql_query_derive = { path = "../graphql_query_derive", version = "0.8.0" }
-serde = { version = "^1.0.78", features = ["derive"] }
-serde_json = "1.0"
-doc-comment = "0.3.1"
+serde = { version = "^1.0", features = ["derive"] }
+serde_json = "^1.0"
+doc-comment = "^0.3"
 
 [dependencies.futures]
-version = "0.1"
+version = "^0.1"
 optional = true
 
 [dependencies.js-sys]
-version = "0.3.5"
+version = "^0.3"
 optional = true
 
 [dependencies.log]
-version = "0.4.6"
+version = "^0.4"
 optional = true
 
 [dependencies.web-sys]
-version = "0.3.2"
+version = "^0.3"
 optional = true
 features = [
     "Headers",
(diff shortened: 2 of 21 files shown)


tomhoule:
I fixed the merge conflicts and already merged the commit about renaming rust keywords (thanks!).

Next step is to look at this PR and https://github.com/graphql-rust/graphql-client/pull/257 to figure out the best way to normalize the generated query struct names when generating from the CLI.

tomhoule:
https://github.com/graphql-rust/graphql-client/pull/257 was merged, so since the rust keywords renaming is already merged, I think we can close this PR :)

PR #260, opened 2019-08-27
Add constructor method "new" for Input types

davidgraeff:
Depends on #253
Depends on #258
Depends on #259

Add constructor method "new" for Input types.
    
The constructor method provides a simpler way to initialize a type with
many or only optional fields. Assuming an input type like this:

```graphql
    input CustomerInput {
      company: String
      customFields: [CustomFieldInput!]
      email: String
      firstName: String
      lastName: String
      phoneNumber: String
    }
```

 The generated code looks like so:

```rust    
    pub struct CustomerInput {
      pub company: Option<String>,
      ...
    }
    impl CustomerInput {
      pub fn new() -> Self {
        Self { company: None, custom_fields: None, ... }
      }
    }
```

And with mandatory fields it looks like this:

```graphql
    input CustomerInput {
      firstName: String
      lastName: String!
    }
```

```rust    
    pub struct CustomerInput {
      pub firstName: Option<String>,
     pub lastName: String,
    }
    impl CustomerInput {
      pub fn new(lastName: String) -> Self {
        Self { firstName: None, lastName: lastName }
      }
    }
```

Because not used code is removed by the linker, those extra generated methods do not add to the final binary size if not used.

## Usecase

We can't derive from `Default` for input types, because they potentially contain mandatory fields like enums not derived from default themselves. An input type with many optional fields therefore requires to name every single field, even if it is just for assigning `None`. With this PR however I can do the following:

```rust
let query = CreateCustomer::build_query(create_customer::Variables {
    customer: create_customer::CustomerInput {
        first_name: Some("first".to_owned()),
        last_name: Some("last".to_owned()),
        email: Some("test@abc.de".to_owned()),
        ..create_customer::CustomerInput::new()
    },
});
```

### DIFF
--- a/examples/github/Cargo.toml
+++ b/examples/github/Cargo.toml
@@ -7,16 +7,16 @@ edition = "2018"
 [dependencies]
 failure = "*"
 graphql_client = { path = "../../graphql_client" }
-serde = "1.0"
-serde_derive = "1.0"
-serde_json = "1.0"
-reqwest = "^0.9.0"
-prettytable-rs = "0.7.0"
-structopt = "0.2.10"
-dotenv = "0.13.0"
-envy = "0.3.2"
-log = "0.4.3"
-env_logger = "0.5.10"
+serde = "^1.0"
+serde_derive = "^1.0"
+serde_json = "^1.0"
+reqwest = "^0.9"
+prettytable-rs = "^0.7"
+structopt = "^0.2"
+dotenv = "^0.13"
+envy = "^0.3"
+log = "^0.4"
+env_logger = "^0.5"
 
 [workspace]
 members = ["."]
--- a/graphql_client/Cargo.toml
+++ b/graphql_client/Cargo.toml
@@ -10,26 +10,26 @@ categories = ["network-programming", "web-programming", "wasm"]
 edition = "2018"
 
 [dependencies]
-failure = "0.1"
+failure = "^0.1"
 graphql_query_derive = { path = "../graphql_query_derive", version = "0.8.0" }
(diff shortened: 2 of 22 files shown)


tomhoule:
Will review this next time I get a chance to work on this project. Sorry for the wait!

mathstuf:
Could you please rebase on `master` for the conflict resolution and CI fixes? Thanks!

mathstuf:
Hmm. Maybe we can somehow divine that `Default` could be implemented for the `Variables` struct? Maybe this is better part of adding separate derives for input and results? That seems more flexible to me personally because it would only apply where it matters rather than making new enum types and such. Thoughts?

davidgraeff:
I will rebase when I find time, sure.

> Hmm. Maybe we can somehow divine that Default could be implemented for the Variables struct? 

I think I don't get you. Because of possible mandatory fields in `Variables`, that the API user **should** set in contrast to relying on default created values.

> Maybe this is better part of adding separate derives for input and results? That seems more flexible to me personally because it would only apply where it matters ....

I'm not against a separation for input and results. But it doesn't change the fact outlined above, does it?

> rather than making new enum types and such. Thoughts?

In this PR at least no new types are created. Only an `impl` is generated.

mathstuf:
That's why I said "divine". We can only do so when there are actually no required fields (`String` has a `impl Default`, but is probably not too relevant here.

The separation would allow you to say `variables_derives = "Debug, Default"` and you just get it for free. If `Default` isn't implementable, the compiler tells you. If it is, you get it (though you can add it for things like `String`-using variables structures, this might actually be wanted).

Once it's rebased, it should be easier to review.

PR #262, opened 2019-08-27
Don't serialize Option::None for Inputs

davidgraeff:
I have noticed that currently the library does not configure serde to skip Option::None and send json `null` values instead.

I'm not sure if it violates the graphql spec (https://graphql.github.io/graphql-spec/draft/#sec-Type-System.Non-Null.Nullable-vs-Optional), but not sending those fields is definitely allowed:

> Fields are always optional within the context of a query, a field may be omitted and the query is still valid. [...] Inputs (such as field arguments), are always optional by default. 

My graphql server however is absolutely not happy about `null` values for optional fields.
This PR adds the serde attribute to not serialize Option::None.

Requires all my other PRs for the sake of avoiding branch chaos on my side. If the last commit is fine, I'll be happy to extract and rebase it on master.

### DIFF
--- a/examples/github/Cargo.toml
+++ b/examples/github/Cargo.toml
@@ -7,16 +7,16 @@ edition = "2018"
 [dependencies]
 failure = "*"
 graphql_client = { path = "../../graphql_client" }
-serde = "1.0"
-serde_derive = "1.0"
-serde_json = "1.0"
-reqwest = "^0.9.0"
-prettytable-rs = "0.7.0"
-structopt = "0.2.10"
-dotenv = "0.13.0"
-envy = "0.3.2"
-log = "0.4.3"
-env_logger = "0.5.10"
+serde = "^1.0"
+serde_derive = "^1.0"
+serde_json = "^1.0"
+reqwest = "^0.9"
+prettytable-rs = "^0.7"
+structopt = "^0.2"
+dotenv = "^0.13"
+envy = "^0.3"
+log = "^0.4"
+env_logger = "^0.5"
 
 [workspace]
 members = ["."]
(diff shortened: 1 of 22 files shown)


tomhoule:
Thanks for noticing that, it's indeed a problem! I have about 30 seconds to think about it now so I'll come back to this later, but I can think of cases where we would want the field to be omitted, and other cases where we would want it to be null. In graphql-ruby for example it's significant (server-side). Maybe `Option<Option<Ty>>`, where `Some(None)` is `null`, and just `None` would omit the field?

theduke:
This is also a problem in Juniper. 
A lot of GraphQL uses the distinction of "absent" vs "null" vs "present" to represent different states. 
EG a `update` resolver could interpret "absent" as ignored, "present but null" to mean: set to null, and "present" to set the value.

Serde can not interpret this out of the box, eg `Option<Option<i32>>` is not supported (it would always be either `None` or `Some(Some(10))`. 

We'd need a enum with a custom `Serialize`/`Deserialize` impl.

```rust
enum Optional<T> {
  None,
  Null,
  Some<T>,
}
```

If the client does get support for this I hope we can coordinate and use the same type.

davidgraeff:
An own enum type sounds like the correct solution. If "from" and "into" traits are implemented, it might not even break existing user code.

mathstuf:
Could you please rebase on `master` to handle conflicts? CI should also be working there.

ronanyeah:
This would actually solve the problem I was trying to articulate here: https://github.com/graphql-rust/graphql-client/issues/364

keeslinp:
I also have a similar issue since omitting a field and passing a null value are not the same. Is there a workaround in the meantime?

John0x:
Why was this just closed without any comment? @tomhoule 
This would have solved the issue that I'm having right now.

mathstuf:
It looks like it was targeting `master` and now the repo uses the `main` branch. Github closed it because it was targeting "nowhere". I don't know if you can update the target branch or not.

Issue #275, opened 2019-09-30
Union splays do not verify that matches are valid


mathstuf:
Given this schema:

```graphql
union SomeUnion = Int | String

type SomeNonUnionType {
    field: Int
}

type Query {
    someField: SomeUnion
}
```

This query has no diagnostics:

```graphql
query {
    someField {
        __typename
        ... on SomeNonUnionType {
            field
        }
    }
}
```

(Note that this is trimmed down by hand from my case from the GitHub schema; I'll test its actual behavior later today.)

Inside of a union query, any `on SomeType` should be verified to be a valid union type member. This probably also applies to interfaces.

mathstuf:
Indeed. Using the latest release, this is still not warned about. Looking into making a PR.

tomhoule:
In general it's a known issue that validation isn't complete. There are plans to extract the query validation from Juniper, long term. That would still be a valuable PR if you want to contribute this :)

mathstuf:
On September 30, 2019 5:11:54 PM EDT, "Tom Houlé" <notifications@github.com> wrote:
>In general it's a known issue that validation isn't complete. There are
>plans to extract the query validation from Juniper, long term. That
>would still be a valuable PR if you want to contribute this :)

Already done :) . We found the error in production because GitHub rejected our queries. I'll add some tests to the PR when I get a chance.

PR #276, opened 2019-09-30
unions: verify that selections are on union variants

mathstuf:
Fixes: #275

---
Interfaces should probably also check that the matched types implement the interface. Looking into that next.


mathstuf:
Now has a test for unions. Interface tests are not complete anyways, so…I'll leave that for now. The logic is analogous anyways.

PR #279, opened 2019-10-02
Structopt 0.3

mathstuf:
---
Everything seems to still work. Just went through [the changelog](https://github.com/TeXitoi/structopt/blob/master/CHANGELOG.md) and updated what looked necessary.


tomhoule:
Nice! thanks!

Issue #277, opened 2019-10-01
CI seems to be broken


tomhoule:

mathstuf:
Indeed. I noticed this when I was working on #276, however, I'm seeing a different error than CI (each toolchain actually seems to disagree on what's going wrong :/ ). I'm seeing:

```
thread 'tests::schema_with_keywords_works' panicked at 'Parse keywords schema: ParseError("Parse error at 7:1\nUnexpected `directive[Name]`\nExpected `scalar`, `type`, `interface`, `union`, `enum` or `input`\n")', src/libcore/result.rs:999:5
```

mathstuf:
This can be closed now :) . See the above referenced PRs.

tomhoule:
Indeed :) thanks!

PR #280, opened 2019-10-02
Ci issues

mathstuf:
---
See #277. Leaving it open since the nightly builder setup seems broken in some way not related to the code (and the travis.yml looks OK to me too…).


mathstuf:
Looks like the nightly builder is timing out or something. I'm not going to handle that here since it's more infrastructure than code related.

mathstuf:
Well, this is now working except for the `nightly` issue. I also filed #281 to try and chip away at the time that everything is taking on CI right now.

Issue #281, opened 2019-10-02
CI turn around time is really long


mathstuf:
I think this comes down to there only being 3 tracks right now. It would be faster to do the WASM tests in a separate job pipeline. `cargo fmt`/`prettier` and `cargo clippy` can also be separate jobs as well (`cargo fmt` is fast since there's no download and `cargo clippy` is just doing `cargo check` under the hood). It also highlights their issues rather than putting them in the middle of the gigantic log. I see just `cargo install wasm-bindgen-cli` taking 450 s (almost 8 minutes). This could almost certainly be cached somehow. Caching the `target` directory may also be useful if some good way of fingerprinting the thing can be found.

I think the biggest problem is the `examples` directories starting everything from scratch again. I think it may be useful to fold them into the standard `examples/foo/bin.rs` pattern and add their oddball deps to `dev-dependencies` (with comments as to why they're there). Each one is taking 4+ minutes and there's only 3 right now.

Thoughts?

mathstuf:
Hrm. Examples are not so easy. They'd need to be moved into a crate directly (probably `graphql_client`?) rather than living at the workspace level. See rust-lang/cargo#7467 for a cargo-level solution to the problem.

tomhoule:
I haven't done the work, but I was thinking about single-file examples as most projects seem to have, instead of separate cargo projects, like you said.

It would also be possible to consider switching to circleCI, maybe github actions or azure pipelines.

mathstuf:
I suspect GitHub Actions is unsuitable due to the time limits and such. 5 builds, ~8 min per build means we have 50 PR updates (including the merge). Without caching of things like the `target` directory between builds, I don't know if that is sufficient; maybe it is. I haven't played with Azure since I don't have a Microsoft account and Circle because they require access to all repos in order to work (or did). That's the main reason I ended up at Cirrus, but whatever you/the org is OK with using/investigating is fine with me.

h-michael:
@mathstuf @tomhoule 
I've used both AzurePipeline and CircleCI in a project, can I try them once?

h-michael:
I'm sorry, I was mistaken. The repository setting can only be changed by the owner. :p

mathstuf:
On November 10, 2019 11:58:10 PM EST, Hirokazu Hata <notifications@github.com> wrote:
>I'm sorry, I was mistaken. The repository setting can only be changed
>by the owner. :p

You should be able to experiment in your fork at least.

[...]

PR #282, opened 2019-10-02
Ci perf

mathstuf:
---
Builds on #280. See #281.


mathstuf:
Heh. Looks like even the ["do more than it should" buildset](https://travis-ci.org/graphql-rust/graphql-client/builds/592379745) is about 5 minutes faster. Cutting out the excess 3 builds should help that (since we're a bit over the concurrency limit Travis allows right now).

However, I suspect that much more improvement could be done. Doing `rustfmt` on [this crate](https://github.com/mathstuf/rust-keyutils) is under 45 s. Granted, it is a much smaller project, but `cargo fmt` shouldn't be taking 8 minutes (most of this is in cache extraction/collection :/ ). I don't know how easy it'd be to run the WASM tests in a Cirrus CI pipeline (probably not too bad), but since the matrix isn't so conflated, the caching may just be better there anyways. It also would be unlikely to have the `if` guards on every single line in the script since the WASM, style, and clippy checks would each have their own job sections. Thoughts on trying Cirrus CI out? GitHub Actions also exists these days, but 2000 minutes may burn quickly on this project without inter-build caching of some kind (currently planned, but not done yet). Or maybe just ditching the Rust "template" stuff Travis has which seems…unfortunately slow.

mathstuf:
So taking into account the staggering that happened here, this took about 23 minutes wall clock time to complete which is a little over half the old wall clock time. There should be some overlap possible between concurrent PRs as well since rustfmt and WASM seem to be faster now (though the wasm-bindgen-cli build time is still ridiculous and 80+% of that job's time).

I suspect the #277 problem with nightly was a busted cache tarball taking 10+ minutes to extract and Travis killing it when the extraction wasn't verbose. I don't know if there's some way you can dump that cache bundle from an admin panel or something on the site if something like that happens again.

mathstuf:
So this is *much* faster than before. Another 50% or so wall-clock time improvement, so it's around 10 minutes now.

But, something changed somewhere that isn't downcasing all of the member fields properly. Time to go through dependencies and find the culprit…

mathstuf:
OK, let's see what kinds of times we can get now that this is based on #284.

mathstuf:
Whew! All builds are under 10 minutes…except WASM which spends most of that time building `wasm-bindgen-cli`. Finding a way to cache that tool build is the next lowest-hanging fruit.

tomhoule:
Awesome!

For wasm-bindgen-cli, there are binary releases here: https://github.com/rustwasm/wasm-bindgen/releases - maybe we could use that? (let's keep it for a separate PR if it's too much to take on in this one)

mathstuf:
Ooh, that would help a lot. I'd rather see it be faster here in one fell swoop :) . Update incoming shortly.

mathstuf:
Well, it seems to be around 9.5 minutes now. About 6 minutes of cache manipulation for the main builders which is…ridiculous IMO. But investigating other CI can be done in #282.

PR #284, opened 2019-10-05
Normalization as attr

mathstuf:
---
Based on #280 so that CI can hope to pass. Does not include #282 since that will use this to improve performance even more.


mathstuf:
Just the nightly cache extraction failure again.

PR #285, opened 2019-10-13
codegen: support distinct response and variable derives

mathstuf:
Instead of having just `additional_derives`, split them apart from each
other. The CLI interface also now follows this pattern.

Note: No backwards compatibility is provided in this commit.

---
See #248 #249 and #260.

This should allow one to set `response_derives` and `variables_derives` separately. For #260, adding `Default` should make it easier to handle all-optional input structures.

Cc: @tomhoule @davidgraeff @mikailbag

### DIFF
--- a/graphql_client/tests/default.rs
+++ b/graphql_client/tests/default.rs
@@ -1,0 +1,14 @@
+use graphql_client::*;
+
+#[derive(GraphQLQuery)]
+#[graphql(
+    query_path = "tests/default/query.graphql",
+    schema_path = "tests/default/schema.graphql",
+    variables_derives = "Default"
+)]
+struct OptQuery;
+
+#[test]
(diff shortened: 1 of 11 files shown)


mathstuf:
@davidgraeff Now with a test that `Default` can be derived for `Variables` structures.

graphql_client/tests/default.rs
```diff
@@ -0,0 +1,14 @@
...
+    query_path = "tests/default/query.graphql",
+    schema_path = "tests/default/schema.graphql",
+    variables_derives = "Default"
+)]
+struct OptQuery;
+
+#[test]
+fn variables_can_derive_default() {
+    let _: <OptQuery as GraphQLQuery>::Variables = Default::default();
+}
```
tomhoule: neat!

PR #288, opened 2019-11-10
Fragment chains

mathstuf:
---
Cc: @cairomassimo @tomhoule


mathstuf:
This handles the test case as reported. I suspect there's probably some other way to "hide" items from the "requires" sweep logic. I'll try to make more test cases.

mathstuf:
Needs #289 to be green. Will rebase once it is merged.

PR #289, opened 2019-11-10
Ci fixes

mathstuf:
---
Looks like nightly schedules would be nice to catch these kinds of things ahead of time.

Cc: @tomhoule

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,10 @@ before_script:
   - if [ "$GRAPHQL_STYLE" = "true" ]; then (npm i -g prettier) fi
   - if [ "$GRAPHQL_STYLE" = "true" ]; then (rustup component add rustfmt clippy) fi
   - if [ "$GRAPHQL_WASM" = "true" ]; then (rustup target add wasm32-unknown-unknown) fi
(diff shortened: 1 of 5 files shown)


tomhoule:
I haven't looked into it in detail (I'm in a bit of a rush today) but it looks like only the wasm and rustfmt builds are failing. We'll revisit the wasm/web support soon and we can look at rustfmt at the same time, so you don't have to fix it if you don't have the time. Should I merge?

mathstuf:
It's the clippy checks that are failing. Will amend. The wasm failures can be more easily ignored for now.

mathstuf:
I'll also note that there are better patterns for enforcing formatting of code (namely automatically fixing the problems instead of just informing). You can read some of the details on that process here:

  - CMake instructions using it: https://gitlab.kitware.com/cmake/cmake/blob/master/Help/dev/review.rst#id5
  - A project that will be migrating to it "soon": https://discourse.vtk.org/t/clang-format-discussion/2073

I'm working on a command line variant of it [here](https://gitlab.kitware.com/utils/rust-ghostflow/merge_requests/206) so that it can be used in GitHub Actions.

tomhoule:
I'm very open to doing that in CI!

mathstuf:
Woo! Green.

PR #292, opened 2019-11-17
Use github actions for CI

h-michael:
The purpose of this PR is only to allow github actions to do the same thing as the check done by travis so as not to spend time on the review.

The following will be done after this PR:
- Use custom docker container
- Use wasm-pack for wasm test
- Windows

ref: #281
You can see this github actions result [here](https://github.com/h-michael/graphql-client/commit/ad858fde701c9062d2f9a8eae3ca3dca87960f7e/checks?check_suite_id=314943311)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,77 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron:  '0 2 * * *'
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        rust: ["stable", "beta", "nightly"]
+        os: [ubuntu-18.04, macos-latest]
+    name: Cargo test
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Checkout sources
+      uses: actions/checkout@v1
+    - name: Install toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: ${{ matrix.rust }}
+        override: true
+    - name: Execute cargo test
+      run: cargo test --all --tests --examples
+  wasm_test:
+    name: Cargo test for wasm
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout sources
+      uses: actions/checkout@v1
+    - name: Install toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+    - name: Setup wasm-bindgen
+      run: |
+        rustup target add wasm32-unknown-unknown &&
+        sudo apt update && sudo apt install -y firefox-geckodriver &&
+        wget https://github.com/rustwasm/wasm-bindgen/releases/download/0.2.54/wasm-bindgen-0.2.54-x86_64-unknown-linux-musl.tar.gz &&
+        mkdir wasm-bindgen &&
+        tar -xzf wasm-bindgen-0.2.54-x86_64-unknown-linux-musl.tar.gz -C wasm-bindgen
+    - name: Execute cargo test
+      run: |
+        export PATH=$PATH:$PWD/wasm-bindgen/wasm-bindgen-0.2.54-x86_64-unknown-linux-musl &&
+        xvfb-run cargo test --manifest-path=./graphql_client/Cargo.toml --features="web" --target wasm32-unknown-unknown
+  lint:
+    name: Rustfmt and Clippy
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout sources
+      uses: actions/checkout@v1
+    - name: Install toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+        components: clippy,rustfmt
+    - name: Execute cargo fmt
+      run: cargo fmt --all -- --check
+    - name: Execute cargo clippy
+      run: cargo clippy -- -D warnings
+  prettier:
+    name: Check prettier
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout sources
+      uses: actions/checkout@v1
+    - name: Install node.js
+      run: sudo apt update && sudo apt-get install -y nodejs
+    - name: Execute prettier
+      run: npx prettier --debug-check -l './**/*.json' './**/*.graphql'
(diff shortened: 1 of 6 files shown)


h-michael:
@mathstuf Would you review this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+    - name: Setup wasm-bindgen
+      run: |
+        rustup target add wasm32-unknown-unknown &&
+        sudo apt update && sudo apt install -y firefox-geckodriver &&
+        wget https://github.com/rustwasm/wasm-bindgen/releases/download/0.2.54/wasm-bindgen-0.2.54-x86_64-unknown-linux-musl.tar.gz &&
```
mathstuf: We should add a note to the relevant `Cargo.toml` that when updating `wasm-bindgen` in the main crate, CI needs updated as well.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+        components: clippy,rustfmt
+    - name: Execute cargo fmt
+      run: cargo fmt --all -- --check
+    - name: Execute cargo clippy
+      run: cargo clippy -- -D warnings
```
mathstuf: Add `--tests --examples` as well.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+  push:
+  pull_request:
+  schedule:
+    - cron:  '0 2 * * *'
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        rust: ["stable", "beta", "nightly"]
```
mathstuf: Can we also test the MSRV? Do we even know what it is?

mathstuf:
> Use custom docker container

I don't think this is necessary. The times for installing the deps seem fast enough. Not having to rebuild a Docker image all the time is going to be better for everyone involved :) .

mathstuf:
> -Z minimal-versions builder

This should be tested with the MSRV.

h-michael:
@mathstuf 

> Why is it not run on this PR?

This is my guess, Github Actions doesn't have a setting to enable or disable, so in projects that don't currently have a Github Actions workflow file, it won't run when the first PR other than the owner creates it. I think.

> Caching of some kind would be nice (though I suspect this will have to wait on support from Github).

Github Actions can now use cache. However, this project does not have the Cargo.lock file that is usually specified as the cache key. I have not yet verified whether Cargo.toml is the appropriate key instead.

> Can we also test the MSRV? Do we even know what it is?

AFAK, this project is not currently configured for MSRV. I think this needs to be discussed separately

h-michael:
I'll check tonight because wasm failed to build.

mathstuf:
From the log:

```
  rust wasm file: 0.2.55
  this binary: 0.2.43 (402e72ed3)
```

Maybe we should pin wasm using an `=` dependency instead of letting it float? A link to the "make wasm-bindgen stable" issue (if any exists; I don't see one at a glance) would be great.

As for clippy, it looks like we have a suppression to add or an example to finish:

```
error: unused implementer of `futures::future::Future` that must be used
   --> examples/web/examples/web.rs:146:5
    |
146 |     load_more();
    |     ^^^^^^^^^^^^
    |
    = note: `-D unused-must-use` implied by `-D warnings`
    = note: futures do nothing unless polled
```

examples/web/examples/web.rs
```diff
@@ -143,7 +143,6 @@ pub fn run() {
         .append_child(&message_area)
         .expect_throw("could not append message area");
 
-    load_more();
```
h-michael: `load_more` is called in `add_load_more_button()`, so we can remove this line.

.github/workflows/ci.yml
h-michael: AFAIK, graphgl-client does not yet specify MSRV.

h-michael:
@mathstuf I'm sorry to be late, and it's ready for review.

.travis.yml
```diff
@@ -29,10 +29,7 @@ before_script:
   - if [ "$GRAPHQL_STYLE" = "true" ]; then (npm i -g prettier) fi
   - if [ "$GRAPHQL_STYLE" = "true" ]; then (rustup component add rustfmt clippy) fi
   - if [ "$GRAPHQL_WASM" = "true" ]; then (rustup target add wasm32-unknown-unknown) fi
-  - if [ "$GRAPHQL_WASM" = "true" ]; then (wget https://github.com/rustwasm/wasm-bindgen/releases/download/0.2.54/wasm-bindgen-0.2.54-x86_64-unknown-linux-musl.tar.gz) fi
-  - if [ "$GRAPHQL_WASM" = "true" ]; then (mkdir wasm-bindgen) fi
-  - if [ "$GRAPHQL_WASM" = "true" ]; then (tar -xzf wasm-bindgen-0.2.54-x86_64-unknown-linux-musl.tar.gz -C wasm-bindgen) fi
-  - export PATH=$PATH:$PWD/wasm-bindgen/wasm-bindgen-0.2.54-x86_64-unknown-linux-musl
+  - if [ "$GRAPHQL_WASM" = "true" ]; then (cargo install wasm-bindgen-cli --force) fi
```
mathstuf: The nice bit about the former was that it kept times down. Now it's taking 6 minutes to do the build. It seems to fit in the times other builds are doing though. It'd be nice if there were a download link for the "latest" release. Not something we can fix I suppose.

graphql_client/Cargo.toml
```diff
@@ -51,7 +51,8 @@ optional = true
 reqwest = "^0.9"
 
 [dev-dependencies]
-wasm-bindgen-test = "0.2.43"
+# Note: If we bumpup wasm-bindge-test version, we should change CI setting.
```
mathstuf: Well, CI is now actually OK since we grab the newest version :) .

h-michael:
@mathstuf ty :)
I'll try to use CI cache in next PR.

h-michael:
Now, github actions runs :)
https://github.com/graphql-rust/graphql-client/runs/352069184
https://github.com/graphql-rust/graphql-client/runs/352069164

PR #295, opened 2019-12-20
Use 'actions/cache@v1' for CI

h-michael:


h-michael:
If we want to use cache, we remove Cargo.lock from .gitignore.

PR #296, opened 2019-12-20
Use 'actions/cache@v1' for CI

h-michael:


h-michael:
Waiting resolve [this issue](https://github.com/actions/cache/issues/133).

.github/workflows/ci.yml
```diff
@@ -27,6 +27,14 @@ jobs:
...
         toolchain: ${{ matrix.rust }}
         override: true
+    - name: Execute cargo update for creating Cargo.lock
+      run: cargo update
+    - name: Cache cargo
+      uses: actions/cache@v1
+      with:
+        path: ~/.cargo/
+        key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+      run: sudo chown -R $(whoami):$(id -ng) ~/.cargo/
```
h-michael: ref: https://github.com/actions/cache/issues/133#issuecomment-599102035

.github/workflows/ci.yml
```diff
@@ -27,6 +27,15 @@ jobs:
         profile: minimal
         toolchain: ${{ matrix.rust }}
         override: true
+    - name: Execute cargo update for creating Cargo.lock
+      run: cargo update
+    - name: Change cache permission
+      run: sudo chown -R $(whoami):$(id -ng) ~/.cargo/
```
h-michael: ref: https://github.com/actions/cache/issues/133#issuecomment-599102035

mathstuf:
Hmm. It doesn't look like this saved much time over [a `master` run](https://github.com/graphql-rust/graphql-client/actions/runs/57299709) (6m23s there, 6m52s here). I expect caching to take some time, but it'd be nice to see how much this is actually saving.

I suspect caching the `wasm-bindgen` build is likely to get *much* more out of this since it seems to be the lowest-hanging fruit.

The approach seems good; I'd just like to see how much it actually improves things (maybe it needs merged to show that?).

PR #308, opened 2020-02-05
Refactor schema representation and code generation

tomhoule:
This is a huge refactoring based on an idea to model the schema with a proper normalized representation. The results so far are promising in terms of code clarity and maintainability. At the moment, this is far from complete.

Previously, name resolution, query validation and code generation all happened in one pass. In this PR, these three are cleanly separated, leading to much more understandable code in my opinion.


h-michael:
@tomhoule I think it is better to divide PR into changing the error lib and refactoring, but what do you think?

tomhoule:
I agree in principle. We'll have to ignore all merge conflicts in graphql-client-codegen though, it's nearly  a complete rewrite.

pickfire:
Isn't it bad to be depending on crates such as `failure` and `anyhow` for a library which is designed for application side so user have the ability to opt in to such crates rather that having the crate that they use decide what error handling libraries should they use?

tomhoule:
Agreed that we shouldn't depend on any external crate if we can help it, but `anyhow` is already an improvement over `failure` because it works with `std::error::Error` instead of a custom trait.

It's a bandwidth issue mostly, we could get rid of it if we found time to define good custom error types.

yuyoyuppe:
I've been using this library successfully, thank you folks for it. 

Quick question: will this refactoring cover the case where we use a single fragment in multiple queries and there're multiple incompatible types with identical layout being generated for each query? Perhaps I'm missing an existing feature, but for now, I've just used the following workaround:
```rust
impl From<repo_items::PullRequestView> for single_pr::PullRequestView {
    fn from(view: repo_items::PullRequestView) -> Self {
        let se = serde_json::to_string(&view).unwrap();
        serde_json::from_str(&se).unwrap()
    }
}

impl From<single_pr::PullRequestView> for repo_items::PullRequestView {
    fn from(view: single_pr::PullRequestView) -> Self {
        let se = serde_json::to_string(&view).unwrap();
        serde_json::from_str(&se).unwrap()
    }
}

// etc. for all other fragments
```

I've also considered simply transmuting between the types them, but doing that [is currently UB in Rust](https://github.com/rust-lang/unsafe-code-guidelines/issues/35).

pickfire:
@yuyoyuppe But that seemed bad, it feels like it is wasting resources to serialize and deserialize again to the same thing. Can you do this instead?
```rust
impl From<repo_items::PullRequestView> for single_pr::PullRequestView {
    fn from(view: repo_items::PullRequestView) -> Self {
        Self { ..view }
    }
}

impl From<single_pr::PullRequestView> for repo_items::PullRequestView {
    fn from(view: single_pr::PullRequestView) -> Self {
        Self { ..view }
    }
}
```

yuyoyuppe:
@pickfire 

> But that seemed bad, it is wasting resources to serialize and deserialize again to the same thing.

That was my initial point. The current codegen behavior also forces users to write many boilerplate-heavy impls. Of course I'd like to avoid it.

> Can you do this instead?

[Nightly Rust doesn't support this syntax](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=7c8b24f8d77038927aedafa71870e3b0), is there a feature flag for that? It's certainly possible to manually reassign each and every field, but that would be a pain to support and kinda defeats the purpose of codegen.

> But I think it is better to keep the same struct, why convert it to the same thing?

What do you mean? All my domain-level functions can operate only on a single fragment type, e.g. `repo_items::PullRequestView` _or_ `single_pr::PullRequestView`, unless I'm wrapping everything in a macro.

tomhoule:
The refactoring is purely internal, I'm aiming to support the same features as before at the end. But yes, the fragment sharing problem is one I'd like to solve, but I haven't figured out an API that wouldn't be error prone yet. You would have to specify where to import the fragments from for each query, and then use a separate macro to generate only the fragments somewhere, and you wouldn't get good error messages. One other solution would be generating code for all the operations in a query document at the same time, but it's a big change.

yuyoyuppe:
Assuming the first solution is chosen and regarding the error message quality, otoh, what if we use proc-macro attributes for fragments generation instead of a separate macro? For example in my case:

```rust
#[derive(GraphQLQuery)]
#[graphql(
    schema_path = "submodules/graphql-schema/schema.graphql",
    query_path = "queries.graphql",
    response_derives = "Serialize,Debug,",
    impl_fragments = "PullRequestView",
)]
pub struct RepoItems;

#[derive(GraphQLQuery)]
#[graphql(
    schema_path = "submodules/graphql-schema/schema.graphql",
    query_path = "queries.graphql",
    response_derives = "Serialize,Debug,",
    use_external_fragments = "PullRequestView",
)]
pub struct SinglePR;

```
So we'd have a single implementer macro and `n` "user" macros. Then inside of each fragment macro expansion will be something like:
```rust
use schema::external_fragments::PullRequestView; // did you forget to impl this fragment via impl_fragments?
```
so we'll get a this kind of error:
```rust
   |
15 | use PullRequestView; // did you forget to impl this fragment via impl_fragments?
   |     ^^^^^^^^^^^^^^^ no `PullRequestView` external crate
```

For multiple implementers we'd get a similar ``PullRequestView` redefined here` with struct declaration and a comment. That may be considered a bit hacky, but I believe it'd produce much confusion among the users. 
It would also preserve backward compatibility and allow users to control the behavior further (not sure why they would want it tho :D)

pickfire:
Why doesn't graphql-client take an approach similar to diesel? Do the code generation once and then a minimal code generation later on to just use the generated code?

tomhoule:
> Why doesn't graphql-client take an approach similar to diesel? Do the code generation once and then a minimal code generation later on to just use the generated code?

Diesel only generates code for the schema definition, a graphql client needs both the schema and the query to determine the return types (otherwise it would have its own query builder like diesel).

tomhoule:
Let's open an issue for the fragment sharing thing. Unfortunately I don't think the error messages would as good as you describe, because the proc macro does not know what is in scope or not, it can only generate an import, and the error won't be super user-friendly.

yuyoyuppe:
yeah, I was hoping that even though macros don't understand modules, we could get a error line with a generated comment, but of course you're right and it's won't be possible w/o nightly and some macro backtracing flag.

I've tried adding something like
```rust
use please_impl_external_fragments_for_NonexistentFragment_once::NonexistentFragment;
```
[there](https://github.com/graphql-rust/graphql-client/blob/e1beed8a2f5b8d78b5bacd4477b6aa0e1420b77d/graphql_client_codegen/src/generated_module.rs#L60) to quickly check what kind of error I'd get:

```rust
15 | #[derive(GraphQLQuery)]
   |          ^^^^^^^^^^^^ use of undeclared type or module `please_impl_external_fragments_for_NonexistentFragment_once`

```
not a perfect solution of course, but at least the compilation was aborted and we wouldn't get 50 more errors due missing module type.

Anyway, [I've created an issue](https://github.com/graphql-rust/graphql-client/issues/312) for this, and would like to offer a help implementing it once this refactoring is merged and the interface design for fragment sharing is agreed upon!

tomhoule:
Great, thanks a lot for the feedback and creating the issue :) I hope I can finish this refactoring soon. It will make iterating over the codegen part of the library much easier, the current code is very tangled and hard to modify.

yuyoyuppe:
sure, get well and ping me if you'd like any help.

pickfire:
> Diesel only generates code for the schema definition, a graphql client needs both the schema and the query to determine the return types (otherwise it would have its own query builder like diesel).

Yes, what I meant is to do code generation based on both schema and types, and then use it in the code, similar to diesel. In graphql sense, similar to apollo codegen, which can generate typescript for angular. https://www.apollographql.com/docs/react/development-testing/developer-tooling/#apollo-codegen. I like the idea, but not so much for their generated output.

tomhoule:
This is getting close - all tests are passing now. I need one more cleanup and polishing session and I think it will be finished, and the codegen will be much more understandable, maintainable and extensible than before.

graphql_client/tests/union_query.rs
```diff
@@ -53,26 +62,81 @@ fn fragment_on_union() {
...
+                on: fragment_on_union::FragmentOnUnionNamesOn::Person,
+            },
+            fragment_on_union::FragmentOnUnionNames {
+                names_fragment: fragment_on_union::NamesFragment::Dog(
+                    fragment_on_union::NamesFragmentOnDog {
+                        name: "Laïka".to_string(),
+                    },
+                ),
+                on: fragment_on_union::FragmentOnUnionNamesOn::Dog,
+            },
```
tomhoule: Maybe this should be collapsed. We can remove one level of nesting.

graphql_client/tests/type_refining_fragments.rs
```diff
@@ -29,15 +29,19 @@ fn type_refining_fragment_on_union() {
                 last_name: Some("Lorde".to_string()),
             }),
             query_on_union::QueryOnUnionNames::Dog(query_on_union::QueryOnUnionNamesOnDog {
-                name: "Laïka".to_string(),
+                dog_name: query_on_union::DogName {
+                    name: "Laïka".to_string(),
+                },
```
tomhoule: We should flatten in this case too.

tomhoule:
Still a bunch of todos left, but we are getting there.

tomhoule:
This is finally ready to review merge! I think graphql_client_codegen is a lot easier to understand and extend after this. The very high level overview is two major changes: 1. query resolution, validation and code generation are now distinct phases that are easier to understand in isolation, and 2. Schema and query are represented as flat vecs instead of nested structs to make traversing relations easier.

The clippy part of CI is failing, but it's not something I can reproduce locally. I'll look into that and squash into one or a few commits before merging. 

@h-michael  and @mathstuf  - no pressure to do this, and no hurry, but if you have time and want to, I'd love a review!

graphql_query_derive/src/lib.rs
```diff
@@ -15,35 +15,28 @@ use proc_macro2::TokenStream;
...
-        Err(err) => panic!(
-            "{}",
-            err.iter_chain()
-                .fold(String::new(), |mut acc, item| {
-                    acc.push_str(&format!("{}\n", item));
-                    acc
-                })
-                .trim_end_matches('\n')
-        ),
+        Err(err) => panic!("{:?}", err,),
```
mathstuf: Trailing comma looks weird without being splayed out.

```suggestion
        Err(err) => panic!("{:?}", err),
```

tomhoule:
Sure, I can make a separate PR for anyhow in the other crates (except graphql_client_codegen). It's kind of a trivial change though, and it will be in part overwritten again here, since there will be glue code to interface with `failure` in graphql_client_codegen. The rest is hard to split because it's basically a rewrite of most of graphql client_codegen, but I'll try to think of something :)

mathstuf:
It's not so much that it's trivial, but that it creates lots of noise that makes it hard to track what changed in the meat of the PR :) .

tomhoule:
PR for the error handling library: https://github.com/graphql-rust/graphql-client/pull/319

tomhoule:
I'm going to merge this soon, since it my view it's a clear improvement, and it's blocking me from working on other issues.

mathstuf:
Sounds good to me; I won't block it, but I also haven't gone through with a fine-tooth comb either. The broad strokes I see when skimming look good though.

tomhoule:
Yes it's way too big to expect a thorough review - I wish I'd managed to break it into reviewable parts, but it started as an experiment and was that for a long time, and by the time I found a structure that worked very nicely there were a bunch of improvements and refactorings mixed together. And I don't have the spare time to make it much tidier unfortunately. Thanks for taking a look though :)

PR #310, opened 2020-03-04
ci: upgrade actions/checkout to v2 from v1

h-michael:
ref: https://github.com/actions/checkout#whats-new


h-michael:
@tomhoule we should fix nightly and beta build error :)

tomhoule:
Definitely! Hopefully I'll have a bit of open source time soon to fix it

PR #324, opened 2020-05-16
Drop anyhow

tomhoule:
Motivation:

- It's one dependency fewer
- Using our own error types will let us slowly improve error reporting


tomhoule:
Next after this PR, I want to replace our hacked together custom HTTP client for the browser feature with an external crate, maybe reqwest.

tomhoule:
Will do :)

PR #328, opened 2020-05-20
Allow disabling ssl verification on schema introspection

danielharbor:
### What this PR does

Introduces a flag for disabling SSL verification when using the `introspect-schema` command:

`no_ssl`: set this to disable SSL verification. The default value is false, so if the flag is not passed, SSL verification will be enabled.

### Why
My team recently decided to use `graphql_client_cli` for introspecting our GraphQL services but quickly realized that there was no way to do this on a host where https is enabled, hence this flag.

### Sample Usage:
```bash
$ graphql-client introspect-schema https:/localhost:8080/graphql --no-ssl --header 'Authorization: Basic a5WcefsppefwfeX2pB25Z3JhtleQ='
```

### DIFF
--- a/graphql_client_cli/README.md
+++ b/graphql_client_cli/README.md
@@ -19,7 +19,7 @@ USAGE:
 FLAGS:
     -h, --help       Prints help information
     -V, --version    Prints version information
-        --no_ssl     Set this option to disable ssl certificate verification. Default value is false.
+        --no-ssl     Set this option to disable ssl certificate verification. Default value is false.
                      ssl verification is turned on by default.
 
 OPTIONS:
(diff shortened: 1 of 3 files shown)


danielharbor:
@tomhoule please could you help take a look at this PR? As far as I can tell, the wasm test failure doesn't seem to be due to these changes.

tomhoule:
It's an additional feature to maintain if we want to switch client, but I think it's fine. I'm trying to minimize the scope as much as possible, since as you noticed, I don't have lots of free time for non-work projects at the moment. Thanks for the PR :)

danielharbor:
Thanks @tomhoule 🙂

PR #337, opened 2020-07-30
Update tests for WASM

MidasLamb:
Trying to get everything up and running for WASM.

I can't get `wasm-pack test` to run on my machine so i'm not sure if this will pass CI.

I've left a `todo!()` in the last test, not sure how specific the error should be checked.


MidasLamb:
Changed it to a draft as I can run CI on my own fork to figure everything out

PR #327, opened 2020-05-17
[WIP] [web] Replace custom HTTP client with external crate

tomhoule:
This simplifies things a lot, and we benefit from other people's work on HTTP in the browser.

Current status: the web example works.

PR #338, opened 2020-07-30
General (optional) Client by using reqwest (as developed by @tomhoule)

MidasLamb:
I've been trying to get @tomhoule code working, but I was wondering why the client was only available for WASM. This PR introduces a general usable client locked after the default feature "http".


MidasLamb:
There is still some cleanup to do, but the bulk is done and I think the suggestion this pr is making (including a http client not only for WASM) is ready for discussion

MidasLamb:
@tomhoule, just a friendly reminder, as you said in #327

graphql_client/src/http.rs
```diff
@@ -0,0 +1,95 @@
...
+            .reqwest_client
+            .post(&self.endpoint)
+            .header("Content-Type", "application/json")
+            .body(serde_json::to_string(&Q::build_query(variables)).unwrap())
+            .send()
+            .await?;
+
+        let text_response = reqwest_response.text().await?;
+
+        dbg!(&text_response);
```
tomhoule, line 59: Let's delete this line.

graphql_client/src/http.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Perform a query.
+    ///
+    // Lint disabled: We can pass by value because it's always an empty struct.
+    #[allow(clippy::needless_pass_by_value)]
+    pub async fn call<Q: GraphQLQuery + 'static>(
+        &self,
+        _query: Q,
+        variables: Q::Variables,
+    ) -> Result<crate::Response<Q::ResponseData>, ClientError> {
+        // TODO: remove the unwrap
```
tomhoule, line 46: I think this line is not relevant anymore.

tomhoule (commented):
I just started reviewing, and I think we have to work a tiny bit more on it, there's a lot of TODOs that need to be addressed before it goes to the main branch — like you noted in your previous messages Thanks for your effort!

tomhoule:
Yes, for the general question: let's have a `reqwest` feature and keep it as thin a wrapper around vanilla reqwest as possible, on all platforms reqwest supports (that should be automatic, without `cfg` flagging on the graphql_client side). I also think we can skip testing it, since it's pretty hard to have a real http test setup and we don't have the time resources to maintain it.

tomhoule:
This feature is implemented in a simpler way in https://github.com/graphql-rust/graphql-client/pull/371

Issue #340, opened 2020-08-26
Cohesiveness between GraphQLQuery struct, its Variables and ResponseData


tomasol:
With current APIs it is possible to have mismatch between the struct, `Variables` and `ResponseData` . Example from the README.md:
```rust
    let request_body = UnionQuery::build_query(variables);
    let client = reqwest::Client::new();
    let mut res = client.post("/graphql").json(&request_body).send()?;
    let response_body: Response<union_query::ResponseData> = res.json()?;
```
The cast in the last line can be replaced with any other `ResponseData` without a compilation error.

 I would like to know if there is interest in fixing the issue. I've prototyped a simple trait that solves the issue for me:
```rust
let response_body = UnionQuery::execute(variables, &client)?;
```
The `client` is actually a trait so it is not coupled with `reqwest`. I will add a PR with draft of the proposed API.

tomhoule:
The `ResponseData` is an associated type on the `Query` trait (as well as variales), so the clients can enforce that all three match. Something like https://github.com/graphql-rust/graphql-client/pull/338 seems like the right direction to me. I would avoid a trait for clients, because they can take a bunch of other configuration we cannot foresee (caching, authentication headers, etc.), and that would force us to box all futures.

tomasol:
Thanks for the quick response. I am not familiar with the issue you are mentioning, can you please provide a pointer? Is it only related to async execution?

PR #341, opened 2020-08-26
WIP: Prototype cohesiveness between GraphQLQuery struct, its Variables and ResponseData

tomasol:
Motivation described in issue #340

### DIFF
--- a/examples/github/examples/github.rs
+++ b/examples/github/examples/github.rs
@@ -15,6 +15,15 @@ type URI = String;
 )]
 struct RepoView;
 
+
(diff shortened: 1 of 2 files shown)


pastjean:
I started making something similar: 

https://gist.github.com/pastjean/e25d56c550bf689827d0d75ce684b4ac

This keeps the whole type together so we can link T , T::Variables and T::ResponseData together
Then it's easy to implement a execute method that takes the PreparedQuery<T> and turn it into a T::ResponseData

PR #342, opened 2020-09-30
Feat: allow disabling derive

jkelleyrtp:
The manual generation tool is enough for me, and the derive crate brings in a lot of overhead. This PR allows clearing the derive feature to use statically-generated types.

### DIFF
--- a/graphql_client/Cargo.toml
+++ b/graphql_client/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Tom Houlé <tom@tomhoule.com>"]
 description = "Typed GraphQL requests and responses"
 repository = "https://github.com/graphql-rust/graphql-client"
 license = "Apache-2.0 OR MIT"
-keywords = ["graphql",  "api", "web", "webassembly", "wasm"]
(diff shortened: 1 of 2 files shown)


graphql_client/Cargo.toml
```diff
@@ -56,6 +50,8 @@ reqwest = "^0.9"
 wasm-bindgen-test = "^0.2"
 
 [features]
+default = ["graphql_query_derive"]
+noderive = []
```
tomhoule, line 54: `graphql_client = { default-features = false, features = ["noderive"]` would be the same as  `graphql_client = { default-features = false`, is there any particular reason to include this?

tomhoule (commented):
Thanks for the PR, looks good! Just one question.

graphql_client/Cargo.toml
jkelleyrtp: No reason - I've dropped it now that you point it out.

tomhoule (approved):
Looks good! Thanks

I'll do a release next time I have a few hours to work on graphql_client.

PR #349, opened 2020-11-12
Add use super::* to main code generation function

simonJasonK:
This allows for arbitrary `response_derives` derive macros to be used.


tomhoule:
I'm a bit worried that indiscriminately importing everything could lead to name conflicts, but it's premature to worry and it fixes the issue for you, so let's merge this :)

Thanks for the PR!

closes https://github.com/graphql-rust/graphql-client/issues/348

PR #350, opened 2020-11-12
Use async/await and exclusively futures 0.3.

boxdot:
Also upgrade wasm-bindgen-futures to 0.4 incl. related test package.
Upgrade reqwest to 0.10 and remove unused dependency.

This is a breaking change, so a version bump is also needed.


tomhoule:
Thanks for the PR! Note that the custom web client will disappear after https://github.com/graphql-rust/graphql-client/pull/338, but in the meantime it's a win.

PR #351, opened 2020-11-18
doctest: Fix regression after upgrade of reqwest to 0.10.

siedentop:
This was introduced in #350. Besides this, my thanks for upgrading
reqwest. The old version cost me some time yesterday.

I do not know why CI did not catch this. Maybe something for the maintainers to
take a look at. Or I made a mistake in my setup.

Below is the error I got:

```shell
---- /Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/doc-comment-0.3.3/src/lib.rs - (line 228) stdout ----
error[E0433]: failed to resolve: use of undeclared type or module `reqwest`
  --> /Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/doc-comment-0.3.3/src/lib.rs:245:27
   |
20 |     let client = reqwest::Client::new();
   |                           ^^^^^^ not found in `reqwest`
   |
help: consider importing this struct
   |
3  | use graphql_client::web::Client;
   |

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ A typed GraphQL client library for Rust.
   ```rust
   use graphql_client::{GraphQLQuery, Response};
   use std::error::Error;
+  use reqwest;
 
   #[derive(GraphQLQuery)]
   #[graphql(
@@ -68,14 +69,14 @@ A typed GraphQL client library for Rust.
   )]
   pub struct UnionQuery;
 
-  fn perform_my_query(variables: union_query::Variables) -> Result<(), Box<dyn Error>> {
+  async fn perform_my_query(variables: union_query::Variables) -> Result<(), Box<dyn Error>> {
 
       // this is the important line
       let request_body = UnionQuery::build_query(variables);
 
       let client = reqwest::Client::new();
-      let mut res = client.post("/graphql").json(&request_body).send()?;
-      let response_body: Response<union_query::ResponseData> = res.json()?;
+      let mut res = client.post("/graphql").json(&request_body).send().await?;
+      let response_body: Response<union_query::ResponseData> = res.json().await?;
       println!("{:#?}", response_body);
       Ok(())
   }
--- a/examples/github/examples/github.rs
+++ b/examples/github/examples/github.rs
@@ -51,7 +51,9 @@ fn main() -> Result<(), anyhow::Error> {
         name: name.to_string(),
     });
 
-    let client = reqwest::blocking::Client::new();
+    let client = reqwest::blocking::Client::builder()
+        .user_agent("graphql-rust/0.9.0")
+        .build()?;
 
     let res = client
         .post("https://api.github.com/graphql")
@@ -59,16 +61,10 @@ fn main() -> Result<(), anyhow::Error> {
         .json(&q)
         .send()?;
 
+    res.error_for_status_ref()?;
+
     let response_body: Response<repo_view::ResponseData> = res.json()?;
     info!("{:?}", response_body);
-
-    if let Some(errors) = response_body.errors {
-        println!("there are errors:");
-
-        for error in &errors {
-            println!("{:?}", error);
-        }
-    }
 
     let response_data: repo_view::ResponseData = response_body.data.expect("missing response data");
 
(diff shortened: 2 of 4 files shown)


siedentop:
CC: @boxdot , @tomhoule  (as you were both involved in #350).

siedentop:
Might make sense to provide some CHANGELOG info, how to upgrade to the newest version. 

The diagnostics are pretty good (it says "use await") but it bears repeating.

Upgrading to the newest version of `graphql_client` requires that `send()` and `json()` become `send().await` and `json().await`.

```
error[E0277]: the `?` operator can only be applied to values that implement `Try`
   --> src/main.rs:101:23
    |
101 |           let mut res = client
    |  _______________________^
102 | |             .post("https://api.github.com/graphql")
103 | |             .bearer_auth(&github_api_token)
104 | |             .json(&q)
105 | |             .send()?;
    | |____________________^ the `?` operator cannot be applied to type `impl futures::Future`
    |
    = help: the trait `Try` is not implemented for `impl futures::Future`
    = note: required by `into_result`
help: consider using `.await` here
```

siedentop:
Two more regressions

* Github example does not work anymore. This worked fine on macOS on tag v0.9.0, but not on latest master (9160417f9cdadf84e47bf0a0c888e9f1f9cd537a).
* My personal example fails with a tokio runtime panic. I'm trying to create a minimal example. And while tokio runtime issues are not the purview of this crate, the issue only appears after upgrading graphql-client.

```shell
❯ cargo run --example github graphql-rust/graphql-client

    Updating crates.io index
   Compiling pin-project-internal v1.0.2
   Compiling graphql-introspection-query v0.1.0 (/Users/user/src/tmp/graphql-client/graphql-introspection-query)
   Compiling graphql_client_codegen v0.9.0 (/Users/user/src/tmp/graphql-client/graphql_client_codegen)
   Compiling graphql_query_derive v0.9.0 (/Users/user/src/tmp/graphql-client/graphql_query_derive)
   Compiling pin-project v1.0.2
   Compiling futures-util v0.3.8
   Compiling graphql_client v0.9.0 (/Users/user/src/tmp/graphql-client/graphql_client)
   Compiling h2 v0.2.7
   Compiling hyper v0.13.9
   Compiling hyper-tls v0.4.3
   Compiling reqwest v0.10.8
   Compiling graphql_query_github_example v0.1.0 (/Users/user/src/tmp/graphql-client/examples/github)
    Finished dev [unoptimized + debuginfo] target(s) in 35.09s
     Running `target/debug/examples/github graphql-rust/graphql-client`
Error: error decoding response body: expected value at line 2 column 1

Caused by:
    expected value at line 2 column 1
```

boxdot:
Thank you for investigating the regression!

> Reproducer is here: https://github.com/siedentop/graphql-reproducer

You have a mismatch of `reqwest` version 0.10.8 and `tokio` version 0.3. This version of `reqwest` still runs on tokio 0.2. Changing to tokio 0.2 fixed the panic.

I am still looking into the `expected value at line 2 column 1` problem.

siedentop:
> Changing to tokio 0.2 fixed the panic.
Thank you! This is great.

> I am still looking into the expected value at line 2 column 1 problem.

I found it. It's newline characters in the serialization. 

Repro with curl: `curl -v -H "Authorization: bearer $GITHUB_API_TOKEN" -X POST -d @minimal.broken.json   https://api.github.com/graphql`

```
❯ cat minimal.broken.json
{
  "variables": { "owner": "siedentop", "name": "git-quickfix" },
  "query": "query {   viewer {\\n     login  }\\n}"
}
```

```
❯ cat minimal.json
{
  "variables": { "owner": "siedentop", "name": "git-quickfix" },
  "query": "query {viewer { login  }}"
}
```

siedentop:
Found it. (For real this time ;) )

Put this in: `let body = res.bytes()?; debug!("Bytes: {:#?}", body);`

> Bytes: b"\r\nRequest forbidden by administrative rules. Please make sure your request has a User-Agent header (http://developer.github.com/v3/#user-agent-required). Check https://developer.github.com for other possible causes.\r\n"

siedentop:
@tomhoule , IMO ready to merge.

@boxdot please review.

@tomhoule, one question: Would you be open to a second example very similar to this, that also used tokio to do multiple requests in parallel (while sharing the client connection)?

tomhoule (approved):
lgtm

tomhoule:
@siedentop I would wait for https://github.com/graphql-rust/graphql-client/pull/338 before adding more examples with custom clients. Also I am a bit reluctant to add code when there is so little resources for maintenance. But I would definitely accept a good PR :)

boxdot (approved):

boxdot:
Thank you for fixing the regression introduced by my MR.

PR #354, opened 2020-11-30
Add an option to provide module for the custom scalar definitions

### LINKED ISSUE #245: Force specifying a module for custom scalars
One of the usability issues with the library is that we can't provide good errors for custom scalars (either the types are visible from the generated code, or they're not). In the presence of custom scalars in generated code, we could force the user of the library to specify a module to import the custom scalar types from, as a `derive` attribute.

Feedback on this would be appreciated before me or someone else implements it :)

miterst:
### Description
Currently, the definitions for the custom scalars have to specified in the parent (`super` ) module.
For example if you have a custom `DateTime` scalar it generates for you `type DateTime = super::DateTime;`

This PR adds the option `custom_scalars_module` which can be used to specify the path to the module that contains the definitions for the custom scalars that are necessary for the generated query.

For the above example, setting this option to `crate::custom_scalars` the generated code for the `DateTime` scalar will be 
`type DateTime = crate::custom_scalars::DateTime;`

### How it can be used
 - From the graphql-client
   ```sh
    graphql-client generate query.graphql --schema-path schema.graphql --custom-scalars-module='crate::custom_scalars'
   ```
- deriving
   ```rust
    #[derive(GraphQLQuery)]
    #[graphql(
        schema_path = "schema.graphql",
        query_path = "query.graphql",
        custom_scalars_module = "crate::custom_scalars",
    )]
    pub struct MyQuery;
   ```

Closes #245

### DIFF
--- a/graphql_client_cli/src/generate.rs
+++ b/graphql_client_cli/src/generate.rs
@@ -17,6 +17,7 @@ pub(crate) struct CliCodegenParams {
     pub no_formatting: bool,
     pub module_visibility: Option<String>,
     pub output_directory: Option<PathBuf>,
+    pub custom_scalars_module: Option<String>,
 }
 
 pub(crate) fn generate_code(params: CliCodegenParams) -> Result<()> {
@@ -30,6 +31,7 @@ pub(crate) fn generate_code(params: CliCodegenParams) -> Result<()> {
         query_path,
         schema_path,
         selected_operation,
+        custom_scalars_module,
     } = params;
 
     let deprecation_strategy = deprecation_strategy.as_ref().and_then(|s| s.parse().ok());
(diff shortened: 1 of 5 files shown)


tomhoule:
Amazing, thanks for the PR! I don't have time to review right away, but I will come back to it very soon. Is it already ready for feedback?

miterst:
> Amazing, thanks for the PR! I don't have time to review right away, but I will come back to it very soon. Is it already ready for feedback?

Thank you!
I saw that some tests on the nightly build had failed(passed on my machine :sweat_smile: ).
I will try to fix them later.

tomhoule:
I just did a first read-through over my lunch break, and it looks great. Next steps for me is looking into the CI failures in case you need help, and confirm manually that it works, then I think we can merge :) A release is also long overdue, so I'll try to find time for that after this PR is merged.

miterst:
> I just did a first read-through over my lunch break, and it looks great. Next steps for me is looking into the CI failures in case you need help, and confirm manually that it works, then I think we can merge :) A release is also long overdue, so I'll try to find time for that after this PR is merged.

The tests were failing because there was a compiler bug in Rust unstable and was affecting one of the WASM examples.
With the new Rust unstable it seems to be compiling without any problems.

tomhoule:
I've delayed manual testing for this too long, I'm going to take some time now to do it.

PR #355, opened 2021-01-13
Remove static lifetime from generated code

visd0m:
I was wondering if it was possible to avoid to generate these `'static` lifetimes.

It would avoid to trigger the [clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_static_lifetimes) on generated code.
```
constants have by default a `'static` lifetime
```


tomhoule:
Looks good! Can you check if the clippy error making the build fail is justified? It's the only concern, apart from that, as far as I'm concerned we can merge this PR.

tomhoule:
Everything looks good, thanks for the PR!

Issue #358, opened 2021-03-07
Deserialize GraphQL enum into user defined Rust enum


jakmeier:
Hi,

I have been using this crate for more than a year in the client of a [hobby project browser game](https://github.com/jakmeier/paddlers-browser-game). Love this crate, thanks a bunch for the great work!

One thing has been bothering me for a while. In the code generated for a query, is it possible to define an existing Rust enum to use for deserialization instead of generating a new one? I'm thinking of something akin to custom scalars, where any Rust object with the same name as the GraphQL type defined in the schema can be used, as long as it implements `Deserialize`.

As far as I can tell, nothing like that is supported today, is that correct? I'm happy to dig into the code and see if I could come up with a PR as an implementation suggestion, if the maintainers of this crate think it would be a worthwhile feature to add.

### Motivation
My motivation for this is the following.

I have several Rust enums defined in a crate that I import in the client and the server. The Rust enums are simple C-style enumerations without fields, hence I map them to GraphQL enums.

The server uses Juniper to generate and serve a GraphQL interface. Using the `graphql_client_cli`, I create a `schema.json` from a running backend instance. Then I generate the code for the queries. This generates new enum types for each query.

So far so good. But I want to use the same enum type in the client as used by the backend. Thus, I have to manually translate from the generate enum back to the original enum type. I have to do this for every query where the enum is used. When I add a enum variant, I have to change the translation code for all queries.

Being able to specify to deserialize into my Rust enum directly would save me a lot of boiler-plate code.

### Example
Shared Rust enum used by client and server:
```rust
#[derive(Clone, Copy, Serialize, Deserialize)]
#[cfg_attr(feature = "backend", derive(juniper::GraphQLEnum))]
pub enum BuildingType {
    Tree,
    Temple,
}
```
Schema generated by Juniper:
```json
{
          "description": null,
          "enumValues": [
            {
              "deprecationReason": null,
              "description": null,
              "isDeprecated": false,
              "name": "TREE"
            },
            {
              "deprecationReason": null,
              "description": null,
              "isDeprecated": false,
              "name": "TEMPLE"
            }
          ],
          "fields": null,
          "inputFields": null,
          "interfaces": null,
          "kind": "ENUM",
          "name": "BuildingType",
          "possibleTypes": null
        },
```
Query:
```graphql
query BuildingsQuery($village_id: Int!) {
  village(villageId: $village_id) {
    buildings {
      id
      buildingType
    }
  }
}
```
[...]

PR #359, opened 2021-03-07
Option for externally defined enums

jakmeier:
Allow to deserialize GraphQL enums into user defined Rust enums, as described in #358


jakmeier:
Here is a first suggestion to add the feature requested in #358. The implementation is very simple. Just don't generate enums if it is in the specified list of external enums. Most of the changes are for adding the option and parsing it. Looking forward to your input on this when you have time to take a look.

Example usage:
```
#[derive(GraphQLQuery)]
#[graphql(
    schema_path = "api/schema.json",
    query_path = "api/queries/buildings_query.graphql",
    extern_enums("BuildingType")
)]
pub struct BuildingsQuery;
```
I tested this with my game and 7 queries which use 1-2 enums each. It works as intended for me. As expected, I was able to remove hundreds LOC of boiler-plate code as a result of this. :-)

This is intentionally a minimal change to get the feature done. I guess from my point of view, this is all I really need. But I'm happy to add more bells and whistles, like the option to rename the type name, if that is desired. 

What I'm most uncertain about in the current implementation is the error handling. I tried to keep it in line with existing code. It seems that errors due to incorrectly formatted attributes are ignored silently and compilation proceeds as if the attributes were not defined. This is what the new `extern_enums` attribute does now, too. Let me know if you prefer something else.

Also, please let me know if anything else I should do for the PR, such as writing changelogs or documentation.

graphql_client_codegen/src/codegen/enums.rs
```diff
@@ -23,7 +23,9 @@ pub(super) fn generate_enum_definitions<'a, 'schema: 'a>(
     );
     let normalization = options.normalization();
 
-    all_used_types.enums(query.schema).map(move |(_id, r#enum)| {
+    all_used_types.enums(query.schema)
+        .filter(move |(_id, r#enum)| !options.extern_enums().contains(&r#enum.name))
+        .map(move |(_id, r#enum)| {
```
tomhoule, line 28: I assume this _just works_ because the generated modules already `use super::*`?

tomhoule (approved):
Looks good! A test would be great if you have time for it, otherwise I'll write one next time I have a few hours to work on graphql-client.

graphql_client_codegen/src/codegen/enums.rs
jakmeier: Yes, it very much relies on that. It's simple and it works. :) Unless you think `use super::*` might be removed in the future, I don't really see this as a problem. But happy to look into alternatives if you think there could be a cleaner solution.

jakmeier:
> Looks good! A test would be great if you have time for it, otherwise I'll write one next time I have a few hours to work on graphql-client.

Absolutely, I will create a test shortly

jakmeier:
I added two basic tests.

Checking the CI again, there are no new issues.

However, I just got a bit confused with this error that started failing on master in the [daily run](https://github.com/graphql-rust/graphql-client/actions/runs/626308064) from 3 days ago:
```
 failures:

---- tests::schema_with_keywords_works stdout ----
thread 'tests::schema_with_keywords_works' panicked at 'assertion failed: generated_code.contains(\"extern_\")', graphql_client_codegen/src/tests/mod.rs:32:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrac
```
Really looks like something that I could have caused with `external_enum`! But since master fails with the same error, it can't be my change. I can reproduce it locally when updating to `nightly-2021-03-06` but it was fine before.

tomhoule:
The tests look good! I think we can merge this, I sadly don't have time to invest in looking at what's wrong in CI now, but I definitely will when I have a few hours to spend on graphql-client. For now we can totally merge this, thanks a lot for the idea and the contribution!

PR #363, opened 2021-04-13
Fix unnecessary boxes introduced by 472aad50

### LINKED ISSUE #362: Too much boxing since 472aad50 (#361)
It looks like #361 introduced a bug where it incorrectly considers that types are recursive even though they aren't, if one type appears twice in the tree but it's still a tree. This generates unnecessary `Box`es.

Ten0:
Fixes #362 
Additionally uses `&str`s instead of `String`s in that place.

### DIFF
--- a/graphql_client_codegen/src/schema.rs
+++ b/graphql_client_codegen/src/schema.rs
@@ -400,13 +400,13 @@ impl StoredInputType {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Ten0:
CI fails seem unrelated

tomhoule:
Thanks!

PR #365, opened 2021-05-26
fix: missing keyword rename annotation for input object

flisky:


flisky:
Tests pass in local afer rebasing graphql-rust/graphql-client#366.

(Also find another bug about keyword renaming, have no idea any more left ...)

tomhoule:
Looks good to me, thanks for the PR!

PR #371, opened 2021-06-25
Prepare for release

tomhoule:
This is mostly about reducing the amount of moving parts (dependencies, features) so the library is less effort to maintain. Notably, it replaces the custom web client with reqwest. This is in preparation for a 0.10.0 release. The commits should be reviewable separately. We will still need to update the CHANGELOG before release.

Feedback is appreciated :)

closes https://github.com/graphql-rust/graphql-client/pull/338
closes https://github.com/graphql-rust/graphql-client/pull/327


graphql_client/src/reqwest.rs
```diff
@@ -0,0 +1,15 @@
...
+pub async fn post_graphql<Q: GraphQLQuery, U: reqwest::IntoUrl>(
+    client: &reqwest::Client,
+    url: U,
+    variables: Q::Variables,
+) -> Result<crate::Response<Q::ResponseData>, reqwest::Error> {
+    let body = Q::build_query(variables);
+    let reqwest_response = client.post(url).json(&body).send().await?;
+
+    Ok(reqwest_response.json().await?)
+}
```
tomhoule, line 15: This may be too inflexible. Let's wait for feedback.

tomhoule:
@h-michael  @mathstuf pinging you just so you notice this, in case you want to review (no pressure though).

graphql_client/src/reqwest.rs
```diff
@@ -0,0 +1,29 @@
+//! A concrete client implementation over HTTP with reqwest.
+
+use crate::GraphQLQuery;
+
+/// Use the provided reqwest::Client to post a GraphQL request.
+#[cfg(feature = "reqwest")]
```
tomhoule, line 6: We should document the feature (crate top-level docs and README).

graphql_client/src/reqwest.rs
mathstuf: I'd say so. It's missing error handling at the HTTP level at least. See [this code](https://gitlab.kitware.com/utils/rust-ghostflow/-/blob/master/ghostflow-github/src/client.rs#L310) for how GitHub's GQL API is handled.

graphql_client_cli/src/introspect_schema.rs
```diff
@@ -20,11 +19,11 @@ pub fn introspect_schema(
     authorization: Option<String>,
     headers: Vec<Header>,
     no_ssl: bool,
-) -> anyhow::Result<()> {
+) {
     use std::io::Write;
 
     let out: Box<dyn Write> = match output {
-        Some(path) => Box::new(::std::fs::File::create(path)?),
+        Some(path) => Box::new(::std::fs::File::create(path).unwrap()),
```
mathstuf, line 26: Not sure I like dropping decent error handling along with `anyhow`, but I prefer better error types even in executables, so…

graphql_client_codegen/src/generated_module.rs
```diff
@@ -6,17 +6,23 @@ use crate::{
...
-    operation_name
-)]
+#[derive(Debug)]
 struct OperationNotFound {
     operation_name: String,
 }
 
+impl Display for OperationNotFound {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("Could not find an operation named ")?;
```
mathstuf, line 18: Why not `write!(f, "Could not… {} …in the query document.", self.operation_name)`?

graphql_client_codegen/src/lib.rs
```diff
@@ -137,24 +143,34 @@ pub fn generate_module_token_stream(
...
+            }
+        }
+    }
+}
+
+impl std::error::Error for ReadFileError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            ReadFileError::FileNotFound { path: _, io_error } => Some(io_error),
+            ReadFileError::ReadError { path: _, io_error } => Some(io_error),
```
mathstuf, line 171: This can be collapsed:

```suggestion
            ReadFileError::FileNotFound { io_error, .. }
            | ReadFileError::ReadError { io_error, .. } => Some(io_error),
```

graphql_client_codegen/src/lib.rs
```diff
@@ -137,24 +143,34 @@ pub fn generate_module_token_stream(
...
-    },
+    FileNotFound { path: String, io_error: io::Error },
+    ReadError { path: String, io_error: io::Error },
+}
+
+impl Display for ReadFileError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ReadFileError::FileNotFound { path, .. } => {
+                write!(f, "Could not find file with path: {}\
```
mathstuf, line 156: This is missing a space after the filepath.

```suggestion
                write!(f, "Could not find file with path: {} \
```

mathstuf (changes requested):

tomhoule:
TODO: check why the readme is missing on crates.io

tomhoule:
Thanks for the review, will read tomorrow

graphql_client_codegen/src/generated_module.rs
tomhoule: Definitely an unnecessary micro-optimization, but `write!()` (format macros in generals) expands to more code than it looks like. It can add up to larger binaries and compile times. I just checked on rust-playground:

```rust
use std::fmt::Write;

fn main() {
    let mut out = String::new();
    write!(out, "{} {} {}", ">", "abcd", "<");
}
```

expands to

```rust
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2018::*;
#[macro_use]
extern crate std;
use std::fmt::Write;

fn main() {
    let mut out = String::new();
    out.write_fmt(::core::fmt::Arguments::new_v1(&["", " ", " "],
                                                 &match (&">", &"abcd", &"<")
                                                      {
                                                      (arg0, arg1, arg2) =>
                                                      [::core::fmt::ArgumentV1::new(arg0,
                                                                                    ::core::fmt::Display::fmt),
                                                       ::core::fmt::ArgumentV1::new(arg1,
                                                                                    ::core::fmt::Display::fmt),
                                                       ::core::fmt::ArgumentV1::new(arg2,
                                                                                    ::core::fmt::Display::fmt)],
                                                  }));
}
```

graphql_client/src/reqwest.rs
tomhoule: So basically return the whole response as error in case status != 200? I haven't worked with GraphQL APIs in a long time, but if I remember correctly, APIs weren't consistent on that point. I'll check the graphql spec tomorrow to see if it says anything about it, but I think it's HTTP agnostic.

graphql_client_cli/src/introspect_schema.rs
tomhoule: The problem of error handling here is mostly about displaying useful context when something fails (nothing is handled). In my experience, anyhow isn't much better at this than panics. I'd be ok reverting the anyhow drop for the CLI. For reference, this is the error type I work with most of the time: https://github.com/prisma/prisma-engines/blob/master/migration-engine/connectors/migration-connector/src/error.rs — it's evolved into its current form over time, and it's exactly the right thing for error reporting in prisma.

graphql_client/src/reqwest.rs
tomhoule: I couldn't find anything HTTP-specific in the spec, but there is [something on graphql.org](https://graphql.org/learn/serving-over-http/#response). The status code isn't important, the response should always be a GraphQL response body. I'm still going to try to return more information, and fail in a useful way if the response body isn't a valid GraphQL response.

tomhoule: Looks like the `reqwest::Error` might already give enough context.

tomhoule: (https://docs.rs/reqwest/0.11.4/reqwest/struct.Error.html)

graphql_client_cli/src/introspect_schema.rs
tomhoule: Working on a quick error type for graphql_client_cli

tomhoule: Pushed.

mathstuf (approved):
Things seem better overall, thanks.

PR #373, opened 2021-06-27
Unknow other variant for fragment

visd0m:
Tries to solve the [issue](https://github.com/graphql-rust/graphql-client/issues/372).

Add the option `fragments-other-variant` to allow the generation of an `Unkown` "serde other" variant in the Enum generated by fragments.

The option can be set as macro attributes or by flag if using the [graphql-client-cli](https://github.com/graphql-rust/graphql-client/tree/master/graphql_client_cli).

As default the option is disabled.


tomhoule (commented):
I did a very quick read and it looks like the PR is on the right path :+1:

Sorry for the small merge conflicts because of https://github.com/graphql-rust/graphql-client/pull/371

tomhoule:
I'll try to find time soon for a proper review. Thanks for the PR!

damjack:
Hi @tomhoule, I have a doubt.
Handling the `Enum` types the lib adds `Other(String)` variant as default (no need to set any flags).

What about doing the same thing for the union/interface types making it the default behaviour.
This :point_up: would make it possible to remove the new flag `--fragments-other-variant` and simplify the code.

tomhoule:
I have only a pretty distant memory of what the generated code looks like, but that sounds very reasonable.

visd0m:
@tomhoule I made a second commit which implements the feature as suggested by @damjack.
You can find:
- in the first commit contains the changes proposed to handle the feature through a new option
- in the second commit contains the changes proposed to handle the feature as default behavior, no options needed

In this way it is easy to see the two versions and decide which one to carry on

tomhoule:
Hi @visd0m , I just had a look at both commits, and I think we should have the opt-in option at first (first commit). In general the commit looks good :+1: I'll cut a release now, so we can merge without worrying too much after that and have time to test.

visd0m:
> Hi @visd0m , I just had a look at both commits, and I think we should have the opt-in option at first (first commit). In general the commit looks good 👍 I'll cut a release now, so we can merge without worrying too much after that and have time to test.

@tomhoule I reverted the last commit and restore the first changes.
Just wondering if after the merge we can think about another release 😬 ? (hoping to use this without pointing to master)

tomhoule:
Yes, we can definitely cut another release as soon as I find some time for graphql-client again, hopefully soon. I'd probably take the opportunity to also change graphql_client_codegen to not depend on syn and quote anymore too.

tomhoule:
@visd0m I changed the main branch for the repo from `master` to `main` and that automatically closed the PR. Please re-open the PR to the `main` branch, sorry for the unintended problem.

PR #377, opened 2021-07-11
Document errors when type aliases for custom scalars are missing

### LINKED ISSUE #320: Documentation: When are type aliases required?
For instance, the [github example](https://github.com/graphql-rust/graphql-client/blob/master/examples/github/examples/github.rs#L8) requires `type URI = String;` or else a rather cryptic error like

```
error[E0412]: cannot find type `URI` in module `super`
   |
   | #[derive(GraphQLQuery)]
   |          ^^^^^^^^^^^^ not found in `super`
   |
   = note: possible candidate is found in another module, you can import it into scope:
           crate::repo_view::URI
```

is raised at compile time.  It would be nice if the documentation mentioned this error message somewhere and explained how to solve it by defining an appropriate type alias.

noritada:
closes #320 

This is just a documentation update on custom scalar types.
Thank you very much!

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,19 @@ struct UnionQuery;
 
 ## Custom scalars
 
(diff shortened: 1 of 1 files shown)


tomhoule (approved):
Looks great, thanks!

PR #379, opened 2021-07-28
Derive from both variables and response for enum

### LINKED ISSUE #378: Enum doesn't inherit derived traits of Variables
Generated enum doesn't get the traits passed to variables-derives, but it does for response-derives.

juchiast:
close #378


juchiast:
Can someone take a loot at this?

tomhoule (approved):
Yes that makes sense, @mathstuf do you see any problem coming out of this?

tomhoule:
Merging for now, let's revert later if we discover problems.

PR #381, opened 2021-08-04
Allow to parse camel case input variables from .graphql files and generate snake_case variables in Rust

o0Ignition0o:
This pull request mimics the current codegen's behaviour for output types.

It allows the codegen to generate snake_case variables whatever the case of the .graphql schema variables, which allows variable names to use the rust case.

I've added tests in `input_object_variables_schema.graphql` to make sure it doesn't break the current codegen as well.

Please let me know if there's anything I can do to improve this PR :)

Jeremy


tomhoule (commented):
I don't have time to do a super thorough review now, but just to kick off the discussion: isn't this a breaking change for users currently using input types?

o0Ignition0o:
Hey, thanks a lot for your comment!

I just tried to investigate a bit more, and it turns out the behavior changed between the 0.9 release and 0.10. I first thought it happened between 0.10 and master, and it was an oversight. It turns out it probably wasn't!

Regardless of whether this would be a good or a bad idea, i think this change would be a breaking one, and here's the things IMO that would help figure out the tradeoff:

- A very valid argument to discard this change is that [the graphql specification](https://spec.graphql.org/draft/#sel-GAFNZPABVBZ73D) is clear about names being case sensitive (albeit for the inputs _and_ the outputs), and underscores are significant, which means `myInput`, `my_input` and `myinput` are three different fields.

- The "output variables" [seem to always be in snake_case on the rust side](https://github.com/graphql-rust/graphql-client/blob/40e6f93c976efba4b8ebb2a4dd125a46bf130beb/graphql_client_codegen/src/codegen.rs#L135), regardless of the overall schema definition (which seems to default to snakeCase), which means there somehow is a behavioral asymmetry at the moment.

- The graphql spec is close to JSON, it makes total sense to have most `.graphql` queries, mutation variables and definition using camelCase (which is what most of the tests do, except this exact spot https://github.com/graphql-rust/graphql-client/blob/main/graphql_client/tests/input_object_variables/input_object_variables_schema.graphql#L39 )

I by no means know the codebase and the project's history well enough to bring a better analysis at the moment. 
Since there's probably a good reason behind this change between 0.9 and 0.10, I'll dig deeper and get back to you with more context :)

If you feel like keeping it as is makes more sense, or even changing the variables casing does, feel free to close this PR and/or let me know, I'll gladly give it a try or refactor what is needed :)

Thanks again for your time!

tomhoule:
That refactoring stretched over a way too long time and I guess this got lost :/ I'll do a more thorough review first, but I now think we should merge this PR and do a patch release.

tomhoule (approved):
Let's merge this — I don't have free time for graphql-client at the moment, so I don't know when I'll get around to doing a patch release unfortunately. Help is very much appreciated. Thanks for the PR :)

o0Ignition0o:
Sure no worries, I hope I can help doing triage or whatever sometimes, if I have a bit more time :)

PR #383, opened 2021-08-04
Support Paths in response derives.

o0Ignition0o:
Up until now only [Ident](https://docs.rs/syn/1.0.74/syn/struct.Ident.html) were supported in the response_derives section,
which unfortunately wouldn't allow us to provide fully qualified names.

This commit enables [Path](https://docs.rs/syn/1.0.74/syn/struct.Path.html) support, and provides us with an error message
if we couldn't parse a response_derive as a valid Path.

### DIFF
--- a/graphql_client_codegen/src/codegen.rs
+++ b/graphql_client_codegen/src/codegen.rs
@@ -165,7 +165,10 @@ fn generate_scalar_definitions<'a, 'schema: 'a>(
 }
 
 fn render_derives<'a>(derives: impl Iterator<Item = &'a str>) -> impl quote::ToTokens {
-    let idents = derives.map(|s| Ident::new(s, Span::call_site()));
+    let idents = derives.map(|s| {
+        syn::parse_str::<syn::Path>(s)
+            .expect(format!("couldn't parse {} as a derive Path", s).as_str())
+    });
 
     quote!(#[derive(#(#idents),*)])
 }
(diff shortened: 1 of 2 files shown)


graphql_client_codegen/src/codegen.rs
```diff
@@ -165,7 +165,10 @@ fn generate_scalar_definitions<'a, 'schema: 'a>(
 }
 
 fn render_derives<'a>(derives: impl Iterator<Item = &'a str>) -> impl quote::ToTokens {
-    let idents = derives.map(|s| Ident::new(s, Span::call_site()));
+    let idents = derives.map(|s| {
+        syn::parse_str::<syn::Path>(s)
+            .expect(format!("couldn't parse {} as a derive Path", s).as_str())
```
tomhoule, line 170: This will allocate a string even in the success case.

This could be changed to something like:

```rust
syn::parse_str::<syn::Path>(s).map_err(|_| format!("couldn't parse {} as a derive Path", s).as_str()).unwrap()
```

tomhoule (commented):
One nitpick but otherwise great :+1: Ready to merge with that one change, thanks for the PR!

graphql_client_codegen/src/codegen.rs
o0Ignition0o: Oh indeed, thanks a lot for pointing it out!

tomhoule:
Thanks!

PR #385, opened 2021-08-10
Displays server response when introspection failed

aedm:
When GraphQL introspection fails, the server response may not necessarily be in JSON format. In that case, the CLI failed to parse and display the response.

This PR makes the CLI display the response error message in plain text.

### Example
```
$ graphql-client introspect-schema --authorization <...> https://api.github.com/graphql
```

Result without this PR:
```
Something else happened. Status: 403
Error: error decoding response body: expected value at line 2 column 1
Location: [...]\graphql_client_cli-0.10.0\src\introspect_schema.rs:61:45
```

Result with this PR:
```
Error: HTTP 403 Forbidden: 
Request forbidden by administrative rules. Please make sure your request has a User-Agent header (http://developer.github.com/v3/#user-agent-required). Check https://developer.github.com for other possible causes.
```

### DIFF
--- a/graphql_client_cli/src/introspect_schema.rs
+++ b/graphql_client_cli/src/introspect_schema.rs
@@ -58,7 +58,10 @@ pub fn introspect_schema(
     } else {
         let status = res.status();
         let error_message = match res.text() {
-            Ok(msg) => format!("HTTP {}: {}", status, msg),
+            Ok(msg) => match serde_json::from_str::<serde_json::Value>(&msg) {
+                Ok(json) => format!("HTTP {}\n{}", status, serde_json::to_string_pretty(&json)?),
+                Err(_) => format!("HTTP {}: {}", status, msg),
+            },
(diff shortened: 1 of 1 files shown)


tomhoule:
I think the status codes are implementation-dependent. I haven't spent a lot of time thinking about this, but what do you think of trying to parse the JSON response, and if that fails print the whole response?

aedm:
> I think the status codes are implementation-dependent. I haven't spent a lot of time thinking about this, but what do you think of trying to parse the JSON response, and if that fails print the whole response?

I assume the response JSON is only valuable if there's no error. Trying to parse the error response is certainly possible, but I'm uncertain about its utility.

Do you have a use case in mind where the error response would be processed further?

aedm:
> parse the JSON response, and if that fails print the whole response?

Done, please take a look.

tomhoule (approved):
Looks good, thanks!

PR #390, opened 2021-10-05
Annotate variable struct fields too

dingxiangfei2009:
This PR attempts to properly rename variable struct fields.

Before this PR, variable field names do not carry `#[serde]` annotation that establishes mapping between the safe name and the original schema name. With this PR, the derive macro will attach those annotation labels to renamed fields.

### DIFF
--- a/graphql_client_codegen/src/codegen.rs
+++ b/graphql_client_codegen/src/codegen.rs
@@ -133,11 +133,9 @@ fn generate_variable_struct_field(
     query: &BoundQuery<'_>,
 ) -> TokenStream {
     let snake_case_name = variable.name.to_snake_case();
-    let ident = Ident::new(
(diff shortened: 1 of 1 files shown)


tomhoule:
Thanks! I don't remember the details of the code, so I can't say in detail, but it looks like a good change. Would it be easy to write a test for? That would make sure we don't introduce a regression in the future.

dingxiangfei2009:
@tomhoule Thank you for the suggestion. Proc-macros are rather difficult to test on its own, so I have decided to put a test in `graphql_client`, along with `input_object_variables` tests. I hope that this would make sense.

dingxiangfei2009:
Hi @tomhoule. Do you think the newly added test sufficient for preventing future regression?

PR #396, opened 2021-11-21
Gql parser 0.4

mathstuf:
---
Cc: @samuela

### DIFF
--- a/graphql_client_codegen/Cargo.toml
+++ b/graphql_client_codegen/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2018"
 
 [dependencies]
 graphql-introspection-query = { version = "0.2.0", path = "../graphql-introspection-query" }
-graphql-parser = "^0.2"
+graphql-parser = "0.4"
 heck = "0.4.0"
 lazy_static = "1.3"
 proc-macro2 = { version = "^1.0", features = [] }
--- a/graphql_client_codegen/src/query.rs
+++ b/graphql_client_codegen/src/query.rs
@@ -59,10 +59,13 @@ pub(crate) struct ResolvedFragmentId(u32);
 #[derive(Debug, Clone, Copy)]
 pub(crate) struct VariableId(u32);
 
-pub(crate) fn resolve(
+pub(crate) fn resolve<'doc, T>(
     schema: &Schema,
-    query: &graphql_parser::query::Document,
-) -> Result<Query, QueryValidationError> {
+    query: &graphql_parser::query::Document<'doc, T>,
+) -> Result<Query, QueryValidationError>
+where
+    T: graphql_parser::query::Text<'doc>,
+{
     let mut resolved_query: Query = Default::default();
 
     create_roots(&mut resolved_query, query, schema)?;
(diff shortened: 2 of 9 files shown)


mathstuf:
Also needs this PR now: https://github.com/graphql-rust/graphql-parser/pull/57

graphql_client_codegen/src/query.rs
```diff
@@ -622,17 +647,20 @@ impl UsedTypes {
...
+where
+    T: graphql_parser::query::Text<'doc>,
+{
     for var in variables {
         query.variables.push(ResolvedVariable {
             operation_id,
-            name: var.name.clone(),
-            default: var.default_value.clone(),
+            name: var.name.as_ref().into(),
+            default: var.default_value.as_ref().map(|dflt| dflt.into_static()),
```
mathstuf, line 663: Needs https://github.com/graphql-rust/graphql-parser/pull/57

graphql_client_codegen/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2018"
 
 [dependencies]
 graphql-introspection-query = { version = "0.2.0", path = "../graphql-introspection-query" }
-graphql-parser = "^0.2"
+graphql-parser = "0.3.1"
```
mathstuf, line 12: Needs released. Tracking: https://github.com/graphql-rust/graphql-parser/issues/50

graphql_client_codegen/src/query.rs
samuela: If we start using `'static` things all over when will those values be freed? Does it happen when the corresponding `Value` gets freed? Using `'static` always strikes me as a little sketchy, but it may certainly be worth it here as long as it doesn't leak memory.

mathstuf: It's `'static` because the *type* is not tied to a lifetime since it manages its own memory using `String` instead of borrowing from some other memory that lives…somewhere else. Basically, instead of `Cow<'a, str>` for any `'a` being allowed, we're saying `Cow<'static, str>` which is either a `&'static str` or an owned string. Both know when they need deallocated.

tomhoule:
The extra lifetime and type parameters to do the same thing are a bit of a shame, but it's worth upgrading, thanks for the hard work! The PR looks good :+1: Let's wait for the graphql-parser PR to be merged and merge this.

graphql_client_codegen/src/query.rs
samuela: My understanding is that an object having lifetime `'a` means that it lives at least as long as `'a`. So a lifetime of `'static` implies that the object must live for the lifetime of the program, no? Perhaps I'm missing something...

mathstuf: For values, that's true. For *types*, a lifetime is the maximum time the type is valid for. The [`std::Error::source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) method requires a `'static` bound on its return in order to support downcasting. Since lifetimes don't exist at runtime, the only way you can validly downcast is if the type is not tied to any specific lifetime itself because there's no way to know that some lifetime-bound type is actually living long enough to support the cast.

jmhain:
Is this PR close to being merged? There are a [couple](https://rustsec.org/advisories/RUSTSEC-2019-0036.html) [advisories](https://rustsec.org/advisories/RUSTSEC-2020-0036.html) for `failure` which `graphql_parser` 0.2 depends on.

samuela:
https://github.com/graphql-rust/graphql-parser/pull/57 has been merged for a few months now, so this one should be ready to go but I don't know what the status is

mathstuf:
Sorry for the delay; I've rebased and fix the compilation error I had locally.

mathstuf:
What happened to CI here?

jk-gan:
Hi, may I know what is the status of this PR?

mathstuf:
Waiting on reviewers/maintainers AFAIK.

tomhoule (approved):
I'm not a fan of the extra generics over different string types, but since we depend on graphql parser, let's upgrade. Thanks for the PR!

Issue #367, opened 2021-05-27
Extensions


tculshaw:
Hi, 

First of all, awesome library - thankyou!!

I know this was opened on #89. 

The extensions field IS part of the standard (albeit as an optional field), see `https://github.com/graphql/graphql-spec/blob/main/spec/Section%207%20--%20Response.md` , and (as #89 quoted) "would be easy to implement - maybe as an Option<serde_json::Value>". I think it DOES need to be an Option<serde_json::Value>    

I'm trying to introduce a rate throttle on a Shopify Graphql based workflow which uses the extensions field to show associated costs of calls.

I'm trying to figure out the best way to do this with the existing code (may a trait on the Response?), but I'm a bit of a Rust newbie.

Or is there any plan to add this field directly any time soon? 

Many Thanks,

Tony

mathstuf:
I think adding `extensions` to [this type](https://docs.rs/graphql_client/0.9.0/graphql_client/struct.Response.html) would be the way to go.

PR #399, opened 2021-11-30
Added extension support

pruthvikar:
Added support for extensions as a `serde_json::Value`. This can be further converted into a concrete type with the use of `serde_json::from_value`.

Related: #367

### DIFF
--- a/graphql_client/src/lib.rs
+++ b/graphql_client/src/lib.rs
@@ -290,8 +290,9 @@ pub struct Response<Data> {
     pub data: Option<Data>,
     /// The top-level errors returned by the server.
     pub errors: Option<Vec<Error>>,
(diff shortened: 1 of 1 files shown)


graphql_client/src/lib.rs
```diff
@@ -288,6 +290,8 @@ pub struct Response<Data> {
     pub data: Option<Data>,
     /// The top-level errors returned by the server.
     pub errors: Option<Vec<Error>>,
+    /// The top-level errors returned by the server.
```
tomhoule, line 293: The doc string should probably say something else. Ideally a short explanation with a link to the spec.

tomhoule (commented):
Thanks for the PR! Just one nitpick.

PR #407, opened 2022-01-03
Add feature to use reqwest with rustls rather than native-tls.

qwandor:
Currently, enabling the `reqwest` feature pulls in native-tls. This is annoying in a project that otherwise uses rustls. This change adds a new feature `reqwest-rustls` that also adds the `reqwest` module, but using `reqwest` with rustls rather than native-tls.


tomhoule (approved):
Looks good to me! Have you had a chance to try this branch?

qwandor:
Yes, it works for me: https://github.com/qwandor/octopower/pull/3

tomhoule:
Great, let's merge then :)

PR #409, opened 2022-01-19
Update arg parsing to clap v3

tranzystorek-io:
Updates arg parsing to the recently released `clap` 3 with the `derive` feature (the new `structopt`).

Changes in `clap-derive`:
- Version is no longer propagated to the subcommands by default
- Help output is colored by default


tomhoule (approved):
looks good to me, thanks

PR #415, opened 2022-03-27
Add a directive to suppress all warnings when generating code via the CLI

### LINKED ISSUE #414: Disable warnings and clippy for generated code
The generated code causes a bunch of compiler warnings. It's probably possible to fix these, but the easiest path is just to add `#![allow(clippy::all)]` and `#![allow(warnings)]` at the top of the generated code. I can make a PR for this if this is something you want to do.

autarch:
Fixes #414


graphql_client_cli/src/generate.rs
```diff
@@ -78,7 +80,7 @@ pub(crate) fn generate_code(params: CliCodegenParams) -> CliResult<()> {
     let gen = generate_module_token_stream(query_path.clone(), &schema_path, options)
         .map_err(|err| Error::message(format!("Error generating module code: {}", err)))?;
 
-    let generated_code = gen.to_string();
+    let generated_code = format!("{}\n{}", WARNING_SUPPRESSION, gen.to_string());
```
tomhoule, line 83: We can avoid a large allocation here:

```rust
let generated_code = format!("{WARNING_SUPPRESSION}\n{gen}");
```

should work

tomhoule (changes requested):
Just one nitpick, but lgtm!

tomhoule (approved):

PR #430, opened 2022-08-10
s/Hash/BTree/g in codegen to make the output deterministic

bolinfest:
We were seeing cases where `#[derive(GraphqlQuery)]` was not generating
consistent output. This appeared to be due to the use of `HashMap` and
`HashSet`, whose iteration order is not guaranteed to be consistent.
Switching to `BTreeMap` and `BTreeSet` appears to fix the issue.

### DIFF
--- a/graphql_client/src/lib.rs
+++ b/graphql_client/src/lib.rs
@@ -32,7 +32,7 @@ pub mod reqwest;
 
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-use std::fmt::{self, Display};
+use std::fmt::{self, Display, Write};
 
 /// A convenience trait that can be used to build a GraphQL request body.
 ///
@@ -215,7 +215,7 @@ impl Display for Error {
                 fragments
                     .iter()
                     .fold(String::new(), |mut acc, item| {
-                        acc.push_str(&format!("{}/", item));
+                        let _ = write!(acc, "{}/", item);
                         acc
                     })
                     .trim_end_matches('/')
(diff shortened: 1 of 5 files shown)


LegNeato:
👋. Looks good, running CI. Also, hit me up and let's chat, been a long time!

bolinfest:
@LegNeato indeed, we are long overdue!

I added a commit to the bottom of the stack in this PR to fix the clippy issue caught by CI.

There is also a Prettier issue, but I think the output is also complaining about a bunch of stuff that appears unrelated to my PR.

bolinfest:
Specifically, this runs with an exit code of 1:

```
npx prettier --debug-check -l './graphql_client/**/*.graphql'
```

but based on the output, I'm not clear what it is complaining about.

bolinfest:
FWIW, as best I can tell, the `cargo test` failure was a temporary network error when fetching a dependency?

```
Updating crates.io index
warning: spurious network error (2 tries remaining): SecureTransport error: connection closed via error; class=Net (12)
warning: spurious network error (1 tries remaining): SecureTransport error: connection closed via error; class=Net (12)
error: failed to get `reqwest` as a dependency of package `graphql_client v0.11.0 (/Users/runner/work/graphql-client/graphql-client/graphql_client)`

Caused by:
  failed to load source for dependency `reqwest`

Caused by:
  Unable to update registry `crates-io`

Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  SecureTransport error: connection closed via error; class=Net ([12](https://github.com/graphql-rust/graphql-client/runs/7782827120#step:4:13))
```

PR #437, opened 2022-09-21
Add skip_default_scalars option

### LINKED ISSUE #432: Skip default scalars
Currently, default scalars are hardcoded on the generated code -- Int, Float, String, Boolean, and ID. Requesting a feature to skip defining this scalars.

My particular use case is I wish to override `ID`, since the `ID` I'm receiving follows a certain format which would be better off represented as a custom struct with its own serde serializer, deserializer.

Perhaps introducing a `skip_default_scalars` option makes sense.

fenhl:
Closes #432.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,20 @@ error[E0412]: cannot find type `URI` in module `super`
            crate::repo_view::URI
 ```
 
+If you would like to customize how some of the five default scalars are represented, you can use the `skip_default_scalars` option. You will then have to define your own types for them like for custom scalars.
+
+```rust
+use graphql_client::GraphQLQuery;
+
+#[derive(GraphQLQuery)]
+#[graphql(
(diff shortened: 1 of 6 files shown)


tomhoule (commented):
That looks good to me. It is only missing tests. Would you be prepared to write one? I think we would want to test that including that option, providing custom types for some of the default scalars and deserializing a specific JSON file, we have the values we expect.

mathstuf (changes requested):
Looks good to me; just needs tests.

tomhoule:
This PR is quite old, I'm going to close it, but anybody feel free to reopen if there is interest in this feature, and pushing this to the finish line.

mathstuf:
Can we get a label for "closed for staleness reasons" so that they can be resurrected more easily?

tomhoule:
Done :)

PR #447, opened 2023-04-01
support `extend type Foo`

### LINKED ISSUE #353: Extend keyword?
I have a type such as:

```
type A {
    field_public: Int
}
```

Later on in the same file, I write:

```
extend type A {
   field_private: Int
}
```

It seems this does not get parsed though and the field_private for A is not generated for the code? Does the extend keyword not work?

pizzacat83:
This PR adds support for [the object extension syntax](https://spec.graphql.org/October2021/#sec-Object-Extensions) (`extend type Foo`).

Fixes #353

Other extension syntaxes such as [interface extensions](https://spec.graphql.org/October2021/#sec-Interface-Extensions) are not supported yet.

### DIFF
--- a/graphql_client_codegen/src/schema.rs
+++ b/graphql_client_codegen/src/schema.rs
@@ -327,6 +327,12 @@ impl Schema {
             .expect("Schema::get_object")
     }
 
+    pub(crate) fn get_object_mut(&mut self, object_id: ObjectId) -> &mut StoredObject {
+        self.stored_objects
+            .get_mut(object_id.0 as usize)
+            .expect("Schema::get_object_mut")
+    }
+
     pub(crate) fn get_field(&self, field_id: StoredFieldId) -> &StoredField {
         self.stored_fields.get(field_id.0).unwrap()
     }
--- a/graphql_client_codegen/src/schema/graphql_parser_conversion.rs
+++ b/graphql_client_codegen/src/schema/graphql_parser_conversion.rs
@@ -1,6 +1,8 @@
 use super::{Schema, StoredInputFieldType, TypeId};
 use crate::schema::resolve_field_type;
-use graphql_parser::schema::{self as parser, Definition, Document, TypeDefinition, UnionType};
+use graphql_parser::schema::{
+    self as parser, Definition, Document, TypeDefinition, TypeExtension, UnionType,
+};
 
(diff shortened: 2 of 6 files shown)


pizzacat83:
I apologize for opening PRs with failing checks.

The [lint check](https://github.com/graphql-rust/graphql-client/actions/runs/4737659574/jobs/8532153743?pr=447) seems to be failing due to code unrelated to this PR. Should I change that code in this PR?

tomhoule:
No worries! I'm not working actively on this project, but hopefully someone can do a proper review. Don't worry about clippy if it's unrelated.

mathstuf:
FWIW, the clippy things are fixed in #449.

graphql_client_codegen/src/deprecation.rs
```diff
@@ -8,22 +8,17 @@ pub enum DeprecationStatus {
...
 
 /// The available deprecation strategies.
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Clone, Default)]
 pub enum DeprecationStrategy {
     /// Allow use of deprecated items in queries, and say nothing.
     Allow,
     /// Fail compilation if a deprecated item is used.
+    #[default]
     Deny,
```
mathstuf, line 17: This doesn't preserve the previous value. I think these clippy fixes belong in its own PR to avoid any confusion about what this PR is about.

graphql_client_codegen/src/tests/foobars_query.graphql
```diff
@@ -1,16 +1,16 @@
 query FooBarsQuery {
+  fooBars {
```
mathstuf, line 2: This reindentation is just noise here.

mathstuf (changes requested):

pizzacat83:
Apologies for the noisy commits. I've reverted those changes, so now this PR contains only the relevant diffs.

mathstuf:
Thanks. I've rebased those commits out so that they're not added-and-removed in the same PR. Could you do the same or allow maintainers to push to the branch?

mathstuf:
Oh, it did allow me to do so; the error I saw was about not having permission to push tags.

mathstuf (approved):
I've rebased out the manual fixes and squashed commits together so that it all passes CI in one shot.

PR #449, opened 2023-04-16
gha: only run on the main repo

mathstuf:
Maybe this will screw up PRs though? I don't know, GHA has such poor control here :/ .

See: https://github.com/orgs/community/discussions/9098


tomhoule (approved):
We can try this :+1:  Thanks for the clippy fixes too.

PR #450, opened 2023-05-22
Generate an enum for `@oneOf` input

jbourassa:
The [RFC for `oneOf` on input type](https://github.com/graphql/graphql-spec/pull/825) is currently in at the draft stage. In short, adding the `@oneOf` directive to an input type means _exactly one_ field of the input's field must be provided.

While the spec still hasn't been accepted, there's a strong need for input unions, the spec is pretty far along and already seeing adoption in the community. It's available in the [GraphQL Ruby gem](https://graphql-ruby.org/type_definitions/input_objects.html#oneof), and we make use of it in Shopify Functions.

With this PR, inputs tagged with `@oneOf` will generate a Rust enum instead of a struct.

From the test, this type:
```graphql
input Param @oneOf {
  author: Author
  name: String
  recursiveDirect: Param
  recursiveIndirect: Recursive
  requiredInts: [Int!]
  optionalInts: [Int]
}
```

generates this Rust enum:

```rust
pub enum Param {
    #[serde(rename = "author")]
    Author(Author),
    #[serde(rename = "name")]
    Name(String),
    #[serde(rename = "recursiveDirect")]
    RecursiveDirect(Box<Param>),
    #[serde(rename = "recursiveIndirect")]
    RecursiveIndirect(Box<Recursive>),
    #[serde(rename = "requiredInts")]
    RequiredInts(Vec<Int>),
    #[serde(rename = "optionalInts")]
    OptionalInts(Vec<Option<Int>>),
}
```

That is, it generates an enum with 1 variant per field, where the variant has 1 required field.

-----

Worth noting that we'll only generate an enum when using IDL-defined schema (`.graphql` file), but not when using the introspection JSON response. The reason for that is that introspecting `oneOf` is done through the `__Type.isOneOf` field, which isn't widely implemented. To make it work, we'd have to dynamically select `__Type.isOneOf` in the introspection query, based on user opt-in. Since my use case didn't require it, and the introspection query is currently static, I didn't do it.

### DIFF
--- a/graphql_client_codegen/src/codegen/inputs.rs
+++ b/graphql_client_codegen/src/codegen/inputs.rs
@@ -2,7 +2,7 @@ use super::shared::{field_rename_annotation, keyword_replace};
 use crate::{
     codegen_options::GraphQLClientCodegenOptions,
     query::{BoundQuery, UsedTypes},
-    schema::input_is_recursive_without_indirection,
+    schema::{input_is_recursive_without_indirection, StoredInputType},
 };
 use heck::ToSnakeCase;
 use proc_macro2::{Ident, Span, TokenStream};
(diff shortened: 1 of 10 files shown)


jbourassa:
@surma this is another stab at addressing the same root problem as [Shopify/shopify-function-rust#27](https://github.com/Shopify/shopify-function-rust/pull/27), curious to hear your thoughts.

graphql_client_codegen/src/codegen/inputs.rs
```diff
@@ -17,48 +18,112 @@ pub(super) fn generate_input_object_definitions(
...
+        let name_ident = Ident::new(safe_variant_name.as_ref(), Span::call_site());
+
+        let normalized_field_type_name = options
+            .normalization()
+            .field_type(field_type.id.name(query.schema));
+        let type_name = Ident::new(normalized_field_type_name.as_ref(), Span::call_site());
+
+        // Add the required qualifier so that the variant's field isn't wrapped in Option
+        let mut qualifiers = vec![GraphqlTypeQualifier::Required];
+        qualifiers.extend(field_type.qualifiers.iter().cloned());
```
tomhoule, line 104: That could probably be `qualifiers.extend([GraphqlTypeQualifier::Required].into_iter())`

tomhoule (commented):
I haven't spent time gathering a lot of context or trying the branch, but with that caveat, lgtm :+1:

graphql_client_codegen/src/codegen/inputs.rs
surma, line 104: What about this? Feels slightly more prescriptive :D

```suggestion
        let mut qualifiers = field_type.qualifiers.clone();
        qualifiers.push(GraphqlTypeQualifier::Required);
```

jbourassa: That won't work either for the reason I tried my best to explain in a prior comment:

> * we can't clone and extend `field_type.qualifiers` with `[GraphqlTypeQualifier::Required]`: the required qualifier MUST come first, otherwise it could be interpreted as `field: [Foo!!]` and panics with [double required annotation](https://github.com/graphql-rust/graphql-client/blob/0756a156d123eafd31fff60fa04c84adec22642f/graphql_client_codegen/src/codegen.rs#L210))

If I recall correctly, the qualifiers map this way:
- `Foo`: `[]`
- `Foo!`: `[required]`
- `[Foo!]`: `[list, required]`
- `[Foo!]!`: `[required, list, required]`

surma: I did not realize that order matters. My bad. Yeah let’s just keep as is.

surma (approved):
LGTM!

I pushed a commit to fix the (new?) clippy warnings. If CI passes, we can ship this.

PR #460, opened 2023-11-11
Improve error message when query name is missing

Enselic:
Thanks for this project!

I'm not used to working with GraphQL, so the current short error message confused me. I did not understand what kind of name was missing at first. Eventually I figured it out.

Make the error message more helpful by hinting what the likely (only?) problem is.


TieWay59:
LGTM

tomhoule (approved):

PR #461, opened 2023-11-14
add dependabot config

danieleades:


danieleades:
my motivation for adding this is driven by discovering that my projects which use `graphql-client` end up with multiple versions of a few dependencies. I tracked this down to the fact that this library has a number of out of date dependencies. Dependabot can help keep these up-to-date without additional maintenance burden.

This is also caused somewhat by the slightly strange 'reqwest' feature in this library.

This feature adds a dependency on `reqwest`, but does not re-export the reqwest client. That forces you to depend on `reqwest` transitively through this library and directly from your own. This can lead to multiple versions of reqwest (and all of its dependencies).

In addition to adding dependabot config, i'd suggest re-exporting the reqwest client from the `reqwest` module, or doing a more thorough rework of this part of the library

danieleades:
marking as draft in favour of #465

dependabot should only really be used if an MSRV check is in place to ensure that dependency bumps don't inadvertently bump the toolchain version needed to build the library

danieleades:
no longer required

PR #462, opened 2023-11-15
update ci

danieleades:


TieWay59 (commented):
Donno much about `dtolnay/rust-toolchain` can you tell us some reason?

tomhoule:
actions-rs/toolchain is not maintained and depends on node, while dtolnay/rust-toolchain is bash and maintained, so I support the move.

TieWay59:
I understand `dtolnay/rust-toolchain` now.

BTW replacing `.get(0)` is not related to your ci right? Maybe you need to address it in the PR body or split it in another PR?

danieleades:
> I understand `dtolnay/rust-toolchain` now.
> 
> BTW replacing `.get(0)` is not related to your ci change right? Maybe you need to address it in the PR body or split it in another PR?
> 
> Edit: I understand it's definitely correct & trivial for running the code. But it's a little inconvenient for review when you said A but also done B. People would be confused about whether it's intended.

The change is intentional, because without it the CI jobs for this PR will fail

tomhoule (approved):

tomhoule:
Let's try this.

PR #465, opened 2023-12-02
Add MSRV check to CI

danieleades:
- find MSRV and set in Cargo.toml
- add an MSRV check to CI
- add dependabot config
- commit lockfile to version control for reproducible CI builds


tomhoule:
Hi @danieleades ! I'm going to take on maintainership on this crate again more actively. Do you still want to discuss this? I've always had the mental model that Cargo.lock shouldn't be committed for libraries, but I need to update myself on the current best practices with MSRV.

danieleades:
> Hi @danieleades ! I'm going to take on maintainership on this crate again more actively. Do you still want to discuss this? I've always had the mental model that Cargo.lock shouldn't be committed for libraries, but I need to update myself on the current best practices with MSRV.

there are several reasons to commit the lockfile for a library, some of which are laid out in [this blog post](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html).

Specifically for this project, it means that:

1. different contributors have exactly the same local environment during development
2. the CI environment is using exactly the same dependencies as local development
3. you don't get inadvertent MSRV bumps due to the CI environment resolving newer dependencies than the local lock file
4. cargo-deny results will match the pinned dependencies until the lock file is updated, not just what happened to get resolved at the time the job ran (this library isn't using cargo-deny yet but possibly should)

The only real downside is a bit more noise from dependabot, but i've added a job for automatically merging dependabot PRs for which all the jobs pass (requires enabling 'automerge' for PRs in the repository config)

Take a look at the config and see if it matches the behaviour you would want. For example I could also make minor updates monthly rather than on demand, etc.

tomhoule:
Ah, there's a prettier check failing. I'd be ok with just removing that check for now.

danieleades:
> Ah, there's a prettier check failing. I'd be ok with just removing that check for now.

not related to these changes anyway, i think

danieleades:
@tomhoule i've removed the prettier check

Issue #446, opened 2023-03-27
fragments_other_variant option is always on


frymawe:
The codegen option fragments_other_variant does not create any change in the code generated.  The root cause I ran into was that the schema I was using actually had Other as one of the enumerations causing a conflict with the generated Other(&str).  graphql_client_codegen/src/codegen/enums.rs  Generates the Other(&str) without regard to the settings in options.  It is probably a different issue, but coming up with a name less likely to collide than Other might be a good idea too.

slyons:
I'm getting hit by this as well:
```rust
error[E0428]: the name `Other` is defined multiple times
  --> src/cmds/creds/ls.rs:29:10
   |
29 | #[derive(GraphQLQuery)]
   |          ^^^^^^^^^^^^ `Other` redefined here
```

khumps:
currently running into this as well.

fldu:
problem is still going on.

PR #469, opened 2023-12-21
Add custom name for `Other` variant

slyons:
Addresses #446 and provides a way to specify the name of the fallback variant of the Query enum.

With `serde_enum_fallback_variant_name=Some("Autre")` you now get an enum like so:

```rust
pub enum AnEnum { 
  where_ , 
  self_ , 
  Autre (String) 
}
```


tomhoule:
Hi @slyons, thanks for the PR! I am going to take on maintainership of this crate a bit more actively again, do you still want to take this PR to the finish line? It looks good, I think we should do it.

PR #470, opened 2024-01-04
Support codegen from query string

dphm:
## Context

Our crate generates the GraphQL query that [uses the macro to generate Rust types](https://github.com/Shopify/shopify-function-rust/blob/48167961252422e8e45b43b448dc290d65074450/shopify_function_macro/src/lib.rs#L363-L373). We want to avoid [writing the generated query to a temporary file](https://github.com/Shopify/shopify-function-rust/blob/48167961252422e8e45b43b448dc290d65074450/shopify_function_macro/src/lib.rs#L351).

## Proposal

- Refactor `graphql_client_codegen` and expose a function to generate module token stream from a query string (instead of a query path)
- Maintain backwards-compatibility with the existing `generate_module_token_stream`

### DIFF
--- a/graphql_client_codegen/src/lib.rs
+++ b/graphql_client_codegen/src/lib.rs
@@ -2,13 +2,12 @@
 #![warn(rust_2018_idioms)]
 #![allow(clippy::option_option)]
 
-//! Crate for internal use by other graphql-client crates, for code generation.
-//!
-//! It is not meant to be used directly by users of the library.
+//! Crate for Rust code generation from a GraphQL query, schema, and options.
 
 use lazy_static::*;
 use proc_macro2::TokenStream;
 use quote::*;
+use schema::Schema;
 
 mod codegen;
 mod codegen_options;
@@ -44,67 +43,92 @@ impl std::error::Error for GeneralError {}
 
 type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 type CacheMap<T> = std::sync::Mutex<BTreeMap<std::path::PathBuf, T>>;
+type QueryDocument = graphql_parser::query::Document<'static, String>;
 
 lazy_static! {
-    static ref SCHEMA_CACHE: CacheMap<schema::Schema> = CacheMap::default();
-    static ref QUERY_CACHE: CacheMap<(String, graphql_parser::query::Document<'static, String>)> =
-        CacheMap::default();
+    static ref SCHEMA_CACHE: CacheMap<Schema> = CacheMap::default();
+    static ref QUERY_CACHE: CacheMap<(String, QueryDocument)> = CacheMap::default();
 }
 
-/// Generates Rust code given a query document, a schema and options.
+fn get_set_cached<T: Clone>(
+    cache: &CacheMap<T>,
+    key: &std::path::Path,
+    value_func: impl FnOnce() -> T,
+) -> T {
+    let mut lock = cache.lock().expect("cache is poisoned");
+    lock.entry(key.into()).or_insert_with(value_func).clone()
+}
+
+fn query_document(query_string: &str) -> Result<QueryDocument, BoxError> {
+    let document = graphql_parser::parse_query(query_string)
+        .map_err(|err| GeneralError(format!("Query parser error: {}", err)))?
+        .into_static();
+    Ok(document)
+}
+
+fn get_set_query_from_file(query_path: &std::path::Path) -> (String, QueryDocument) {
+    get_set_cached(&QUERY_CACHE, query_path, || {
+        let query_string = read_file(query_path).unwrap();
+        let query_document = query_document(&query_string).unwrap();
+        (query_string, query_document)
+    })
+}
+
+fn get_set_schema_from_file(schema_path: &std::path::Path) -> Schema {
+    get_set_cached(&SCHEMA_CACHE, schema_path, || {
+        let schema_extension = schema_path
+            .extension()
+            .map(|ext| ext.to_str().expect("Path must be valid UTF-8"))
+            .unwrap_or("<no extension>");
+        let schema_string = read_file(schema_path).unwrap();
+        match schema_extension {
+            "graphql" | "gql" => {
+                let s = graphql_parser::schema::parse_schema::<&str>(&schema_string).map_err(|parser_error| GeneralError(format!("Parser error: {}", parser_error))).unwrap();
+                Schema::from(s)
+            }
+            "json" => {
+                let parsed: graphql_introspection_query::introspection_response::IntrospectionResponse = serde_json::from_str(&schema_string).unwrap();
+                Schema::from(parsed)
+            }
+            extension => panic!("Unsupported extension for the GraphQL schema: {} (only .json and .graphql are supported)", extension)
+        }
+    })
+}
+
+/// Generates Rust code given a path to a query file, a path to a schema file, and options.
 pub fn generate_module_token_stream(
     query_path: std::path::PathBuf,
     schema_path: &std::path::Path,
     options: GraphQLClientCodegenOptions,
 ) -> Result<TokenStream, BoxError> {
-    use std::collections::btree_map;
+    let query = get_set_query_from_file(query_path.as_path());
+    let schema = get_set_schema_from_file(schema_path);
 
-    let schema_extension = schema_path
-        .extension()
-        .and_then(std::ffi::OsStr::to_str)
-        .unwrap_or("INVALID");
-    let schema_string;
+    generate_module_token_stream_inner(&query, &schema, options)
+}
 
-    // Check the schema cache.
-    let schema: schema::Schema = {
-        let mut lock = SCHEMA_CACHE.lock().expect("schema cache is poisoned");
-        match lock.entry(schema_path.to_path_buf()) {
-            btree_map::Entry::Occupied(o) => o.get().clone(),
-            btree_map::Entry::Vacant(v) => {
-                schema_string = read_file(v.key())?;
-                let schema = match schema_extension {
-                    "graphql" | "gql" => {
-                        let s = graphql_parser::schema::parse_schema::<&str>(&schema_string).map_err(|parser_error| GeneralError(format!("Parser error: {}", parser_error)))?;
-                        schema::Schema::from(s)
-                    }
-                    "json" => {
-                        let parsed: graphql_introspection_query::introspection_response::IntrospectionResponse = serde_json::from_str(&schema_string)?;
-                        schema::Schema::from(parsed)
-                    }
-                    extension => return Err(GeneralError(format!("Unsupported extension for the GraphQL schema: {} (only .json and .graphql are supported)", extension)).into())
-                };
+/// Generates Rust code given a query string, a path to a schema file, and options.
+pub fn generate_module_token_stream_from_string(
+    query_string: &str,
+    schema_path: &std::path::Path,
+    options: GraphQLClientCodegenOptions,
+) -> Result<TokenStream, BoxError> {
+    let query = (query_string.to_string(), query_document(query_string)?);
+    let schema = get_set_schema_from_file(schema_path);
 
-                v.insert(schema).clone()
-            }
-        }
-    };
+    generate_module_token_stream_inner(&query, &schema, options)
+}
+
+/// Generates Rust code given a query string and query document, a schema, and options.
+fn generate_module_token_stream_inner(
+    query: &(String, QueryDocument),
+    schema: &Schema,
+    options: GraphQLClientCodegenOptions,
+) -> Result<TokenStream, BoxError> {
+    let (query_string, query_document) = query;
 
     // We need to qualify the query with the path to the crate it is part of
-    let (query_string, query) = {
-        let mut lock = QUERY_CACHE.lock().expect("query cache is poisoned");
-        match lock.entry(query_path) {
-            btree_map::Entry::Occupied(o) => o.get().clone(),
-            btree_map::Entry::Vacant(v) => {
-                let query_string = read_file(v.key())?;
-                let query = graphql_parser::parse_query(&query_string)
-                    .map_err(|err| GeneralError(format!("Query parser error: {}", err)))?
-                    .into_static();
-                v.insert((query_string, query)).clone()
-            }
-        }
-    };
-
-    let query = crate::query::resolve(&schema, &query)?;
+    let query = crate::query::resolve(schema, query_document)?;
 
     // Determine which operation we are generating code for. This will be used in operationName.
     let operations = options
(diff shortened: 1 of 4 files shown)


tomhoule:
I don't have context on the change but that made me think, now that `#[graphql(query = include_str!("./my_query.graphql")]` is possible (pseudo-code, and I haven't checked, but pretty sure it works now), couldn't we make this the one and only way to inject the query? It would be a breaking change, of course, but we can have a deprecation cycle.

dphm:
@tomhoule Oh! I didn't know that existed. I agree that would be a better way to include the query from a file. Do you know if that would still work with all the file path logic (`CARGO_MANIFEST_DIR`)?

Seems like I could still make the changes to accept a query string instead of the file, right?

graphql_client_codegen/src/lib.rs
```diff
@@ -43,83 +44,122 @@ impl Display for GeneralError {
...
 type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
-type CacheMap<T> = std::sync::Mutex<BTreeMap<std::path::PathBuf, T>>;
+type CacheMap<K, V> = std::sync::Mutex<BTreeMap<K, V>>;
+type QueryDocument = graphql_parser::query::Document<'static, String>;
 
 lazy_static! {
-    static ref SCHEMA_CACHE: CacheMap<schema::Schema> = CacheMap::default();
-    static ref QUERY_CACHE: CacheMap<(String, graphql_parser::query::Document<'static, String>)> =
+    static ref SCHEMA_CACHE: CacheMap<std::path::PathBuf, Schema> = CacheMap::default();
+    static ref QUERY_CACHE: CacheMap<std::path::PathBuf, (String, QueryDocument)> =
```
dphm, line 52: Maybe `CacheMap<String, QueryDocument>` if we want to separate caching from reading from query file?

dphm:
For some early 👀 and any thoughts! @surma @nickwesselman @andrewhassan @adampetro

graphql_query_derive/src/lib.rs
```diff
@@ -26,38 +27,48 @@ fn graphql_query_derive_inner(
...
 
-    let query_path = attributes::extract_attr(input, "query_path")?;
-    let query_path = format!("{}/{}", cargo_manifest_dir, query_path);
-    let query_path = Path::new(&query_path).to_path_buf();
-    let schema_path = attributes::extract_attr(input, "schema_path")?;
+    let schema_path = attributes::extract_attr(&ast, "schema_path")?;
     let schema_path = Path::new(&cargo_manifest_dir).join(schema_path);
-    Ok((query_path, schema_path))
+
+    let query_string = attributes::extract_attr(&ast, "query_string").ok();
```
dphm, line 41: Maybe just `query` (as the default or only supported method of providing the query)?

```suggestion
    let query_string = attributes::extract_attr(&ast, "query").ok();
```

graphql_client_codegen/src/lib.rs
```diff
@@ -43,83 +44,122 @@ impl Display for GeneralError {
...
+            "graphql" | "gql" => {
+                let s = graphql_parser::schema::parse_schema::<&str>(&schema_string).map_err(|parser_error| GeneralError(format!("Parser error: {}", parser_error)))?;
+                Schema::from(s)
+            }
+            "json" => {
+                let parsed: graphql_introspection_query::introspection_response::IntrospectionResponse = serde_json::from_str(&schema_string)?;
+                Schema::from(parsed)
             }
+            extension => return Err(GeneralError(format!("Unsupported extension for the GraphQL schema: {} (only .json and .graphql are supported)", extension)).into())
         }
```
dphm, line 108: ❔ Rust Q: Does this value get computed lazily if it's in `{}`?

tomhoule:
> Do you know if that would still work with all the file path logic (CARGO_MANIFEST_DIR)?

The logic is different: the paths in the proc macro are relative to CARGO_MANIFEST_DIR, but `include_str!()` takes a path relative to the file it is located in.

To address the questions in the PR description

> Should backwards-compatibility be maintained, or is a breaking change acceptable? 

It should be backwards compatible at first, unless we decide to make the new attribute argument the only way, but progressively, deprecating the old argument.

graphql_client_codegen/src/lib.rs
```diff
@@ -43,83 +44,122 @@ impl Display for GeneralError {
...
+        (query_string, query_document)
+    })
+}
+
+fn get_set_schema_from_file(schema_path: &std::path::Path) -> Result<Schema, BoxError> {
+    get_set_cached(&SCHEMA_CACHE, schema_path.to_path_buf(), {
+        let schema_extension = schema_path
+            .extension()
+            .and_then(std::ffi::OsStr::to_str)
+            .unwrap_or("INVALID");
```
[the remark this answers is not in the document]
surma: Also I wonder if we should `std::fs::canonicalize` the path to make sure the cache works when there’s multiple Rust files referencing the same schema file.

graphql_client_codegen/src/lib.rs
surma: What was your thinking behind this? The only scenario I can think of is that a developer has two files with identical content and it would allow the crate to skip GQL parsing. I am not sure that scenario is common enough to pay for the complexity. WDYT?

graphql_client_codegen/src/lib.rs
```diff
@@ -43,83 +44,122 @@ impl Display for GeneralError {
...
+) -> Result<(String, QueryDocument), BoxError> {
+    get_set_cached(&QUERY_CACHE, query_path.to_path_buf(), {
+        let query_string = read_file(query_path)?;
+        let query_document = query_document(&query_string)?;
+        (query_string, query_document)
+    })
+}
+
+fn get_set_schema_from_file(schema_path: &std::path::Path) -> Result<Schema, BoxError> {
+    get_set_cached(&SCHEMA_CACHE, schema_path.to_path_buf(), {
```
surma, line 92: Tiny DX hack :D This allows these kind of functions to be called with a `PathBuf`, `&PathBuf`, a `&Path` or even a `&&Path`. It can simplify some of the code further below. Feel free to dismiss this altogether tho.

```suggestion
fn get_set_schema_from_file(schema_path: impl Into<PathBuf>) -> Result<Schema, BoxError> {
    let schema_path = schema_path.into();
     get_set_cached(&SCHEMA_CACHE, schema_path.clone(), {
```

graphql_query_derive/src/lib.rs
surma: I’d vote for not shipping the breaking change just yet, so I think `query_string` for now. If there is a plan for a major release, we can do the refactor for that release.

graphql_query_derive/src/lib.rs
```diff
@@ -26,38 +27,48 @@ fn graphql_query_derive_inner(
...
+            generate_module_token_stream(query_path, &schema_path, options)
+        }
+    };
+
+    module_token_stream.map(Into::into).map_err(|err| {
+        syn::Error::new_spanned(
+            ast,
+            format!("Failed to generate GraphQLQuery impl: {}", err),
+        )
+    })
```
surma, line 66: Mega nit! But I’d argue that this is slightly more idiomatic
```suggestion
    Ok(module_token_stream.map_err(|err| {
        syn::Error::new_spanned(
            ast,
            format!("Failed to generate GraphQLQuery impl: {}", err),
        )?
    }).into()
```

surma (commented):
Great stuff 🎉 Left some comments! Hope that’s helpful.

Would be great to also see a test for `query_string`.

graphql_query_derive/src/lib.rs
dphm: I'm here for the idiomatic nits since I don't write Rust too often. 😃 Thank you!

graphql_client_codegen/src/lib.rs
tomhoule: `PathBuf` implements `Deref<Target = Path>` so except for the owned type, everything should already work.

graphql_client_codegen/src/lib.rs
dphm: Ah okay, I wasn't sure if it was desirable to cache the `QueryDocument` if the string didn't change. I'll remove this if it's only important to track file changes.

tomhoule: The query cache is there because you can have multiple operations (so multiple structs, with multiple derives) for a single GraphQL executable document (= a single file). There are tests for this.

graphql_client_codegen/src/lib.rs
dphm: Oh ahah I meant this part that you addressed in another comment with `value_func`:

> I think this doesn’t actually deduplicate the work, because the function parameter gets evaluated before the function is called. It’s only the insertion that is skipped.

graphql_query_derive/src/lib.rs
dphm: Oh I think this is why the result is being mapped:

```
mismatched types
`proc_macro2::TokenStream` and `proc_macro::TokenStream` have similar names, but are actually distinct typesrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2FUsers%2Fd%2Fsrc%2Fgithub.com%2Fgraphql-rust%2Fgraphql-client%2Fgraphql_query_derive%2Fsrc%2Flib.rs)
lib.rs(27, 6): expected `std::result::Result<proc_macro::TokenStream, syn::Error>` because of return type
lib.rs(185, 1): `proc_macro2::TokenStream` is defined in crate `proc_macro2`
lib.rs(79, 1): `proc_macro::TokenStream` is defined in crate `proc_macro`
lib.rs(61, 5): use `?` to coerce and return an appropriate `Err`, and wrap the resulting value in `Ok` so the expression remains of type `Result`: `Ok(`, `?)`
```

surma: Huh, I thought I tested this locally before commenting this. But again, feel free to ignore.

graphql_client_codegen/src/lib.rs
surma: Ah yeah, good point. I was mostly going by the fact that I saw a `.as_path()` further down.

graphql_client_codegen/src/lib.rs
```diff
@@ -43,83 +44,122 @@ impl Display for GeneralError {
...
+}
+
+fn get_set_query_from_file(
+    query_path: &std::path::Path,
+) -> Result<(String, QueryDocument), BoxError> {
+    get_set_cached(&QUERY_CACHE, query_path.to_path_buf(), {
+        let query_string = read_file(query_path)?;
+        let query_document = query_document(&query_string)?;
+        (query_string, query_document)
+    })
```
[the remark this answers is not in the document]
surma: Oh true. I missed that part. Since this is macro-land, you could prob just unwrap(). But again, not worth blocking this PR on.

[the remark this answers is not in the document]
dphm: What is `move` for here?

graphql_client_codegen/src/lib.rs
dphm: 👍🏽 I passed `schema_path` here and used `key.into()`

dphm:
Ah, I removed the changes to `graphql_query_derive` so I'll move the test out of `graphql_client` to `graphql_client_codegen`

dphm:
> depending on the graphql_client_codegen crate directly

Is this something we want to do?

tomhoule:
> Is this something we want to do?

I can't speak for spotify, but I think that if I had more time to work on this crate, I'd like it to become more modular. The core value proposition is the codegen, and that's just a `fn(graphql_query: &str, graphql_schema: &str, config: Config) -> String` function — we shouldn't force people to use a derive macro to take advantage of it. If it's hard to depend on graphql-client-codegen directly, we could make it easier.

dphm:
https://github.com/graphql-rust/graphql-client/blob/a7283cb25fa332e4fb46f0247db1b173913d629c/graphql_client_codegen/src/lib.rs#L5-L7

Sounds like we should align on a public interface for using the codegen crate directly, and then I can update this comment 😆

dphm:
🤔 We are using the [`derive`](https://github.com/Shopify/shopify-function-rust/blob/48167961252422e8e45b43b448dc290d65074450/shopify_function_macro/src/lib.rs#L364) though, and it would be nice to have the `query` string literal work. I could make that change in a separate PR.

tomhoule:
> Oh, I guess we would replace the entire quoted derive with the call to generate module tokens from graphql_client_codegen.

Yes, that's what I had in mind.

graphql_client_codegen/src/lib.rs
```diff
@@ -44,67 +43,92 @@ impl std::error::Error for GeneralError {}
...
-    };
+    generate_module_token_stream_inner(&query, &schema, &options)
+}
 
-    let query = crate::query::resolve(&schema, &query)?;
+/// Generates Rust code given a query string, a path to a schema file, and options.
+pub fn generate_module_token_stream_from_string(
+    query_string: &str,
+    schema_path: &std::path::Path,
+    options: &GraphQLClientCodegenOptions,
```
dphm, line 114: I suppose I should keep this consistent?

```suggestion
    options: GraphQLClientCodegenOptions,
```

graphql_client_codegen/src/lib.rs
```diff
@@ -44,67 +43,92 @@ impl std::error::Error for GeneralError {}
...
+fn get_set_query_from_file(query_path: &std::path::Path) -> (String, QueryDocument) {
+    get_set_cached(&QUERY_CACHE, query_path, || {
+        let query_string = read_file(query_path).unwrap();
+        let query_document = query_document(&query_string).unwrap();
+        (query_string, query_document)
+    })
+}
+
+fn get_set_schema_from_file(schema_path: &std::path::Path) -> Schema {
+    get_set_cached(&SCHEMA_CACHE, schema_path, move || {
```
dphm, line 78: @surma Do I need this `move`? What about for `get_set_query_from_file`?

surma: Try it without `move`. I did it for clarity (as in: this closure doesn’t mutate external state).

andrewhassan:
Hey @tomhoule @surma! 👋 Would it be possible to get your 👀 on this PR again? Thank you!! 🙏

surma (approved):
LGTM!

graphql_client_codegen/src/lib.rs
```diff
@@ -44,67 +43,92 @@ impl std::error::Error for GeneralError {}
...
+
+fn get_set_schema_from_file(schema_path: &std::path::Path) -> Schema {
+    get_set_cached(&SCHEMA_CACHE, schema_path, || {
+        let schema_extension = schema_path
+            .extension()
+            .map(|ext| ext.to_str().expect("Path must be valid UTF-8"))
+            .unwrap_or("<no extension>");
+        let schema_string = read_file(schema_path).unwrap();
+        match schema_extension {
+            "graphql" | "gql" => {
```
tsetso-spread, line 85: Hello, I think we should also support `graphqls` as it is also a valid schema file extensions

PR #476, opened 2024-03-09
codegen: enable ids to be deserialized from strings or integers

willfindlay:
Although not strictly to spec, some upstream graphql implementations such as the start.gg
API encode their ID types as integers rather than strings. Prior to this commit,
deserialization would fail in such cases, since graphql_client simply type aliases the ID
type to a String. This commit introduces a simple deserialization helper that enables us
to handle both integers and strings while maintaining backward compatbility for downstream
users.

Fixes: https://github.com/graphql-rust/graphql-client/issues/455

### DIFF
--- a/graphql_client/src/lib.rs
+++ b/graphql_client/src/lib.rs
@@ -30,6 +30,8 @@ pub use graphql_query_derive::*;
 ))]
 pub mod reqwest;
 
+pub mod serde_with;
+
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fmt::{self, Display, Write};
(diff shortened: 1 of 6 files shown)


tomhoule:
Hi @willfindlay , thanks for the PR! I am going to return to a bit more active maintainership of this crate. If I understand the implementation correctly, that would be a breaking API change for existing users, right? Could we make the transformation optional?

tomhoule:
Ah wait, it will still be a String in the generated code, so it's not breaking. I think we could merge this.

tomhoule (approved):

PR #480, opened 2024-03-19
support graphqls schema extension

tsetso-spread:
Adding support for `.graphqls` as it is also a valid schema extension


tsetso-spread:
Hey @surma @tomhoule, could you take a look at this, and if good possibly include it in 0.14?

tomhoule (approved):
No objection, seems like a straightforward improvement :+1:

Issue #439, opened 2022-11-01
skip_serializing_none does not work


Nohac:
I tried asking this question on the actual PR https://github.com/graphql-rust/graphql-client/pull/431 that was meant to solve this issue, but got no response, so I'm just gonna move the question into this issue.

I've added `skip_serializing_none` to the struct as stated in the README, but doing so still leaves me with `null` values after serialization.

My first guess was that this hasn't been released yet, so I switch over to the `main` branch in `Cargo.toml`, but that changed nothing. Here's the code I'm using:

```rust
#[derive(GraphQLQuery)]
#[graphql(
    schema_path = "graphql/schema.graphql",
    query_path = "graphql/insert_mutation.graphql",
    response_derives = "Debug",
    skip_serializing_none
)]
pub struct InsertMutation;
```
```rust
let vars = Variables {
    // .. omitted
    user_id: None,
    // .. omitted
};

let q = crate::graphql::InsertMutation::build_query(vars);

let skipped = serde_json::to_string_pretty(&q).unwrap();
println!("{skipped}");
```
Output:
```json
{
  "variables": {
    // .. omitted
    "user_id": null,
    // .. omitted
  },
  ...
}
```

Is there something obvious I'm missing here to make this work, or is this something that warrants it's own issue?

tomhoule:
I unfortunately don't have time to spend on this project to investigate further or fix things, but I'm happy to review PRs. FWIW it looks like a bug.

Nohac:
Thanks for the quick reply.

This is unfortunately a blocking issue for me, unless there's some way to work around this issue. Since hasura 2.x now reports an error if input fields with null values are present, it's no longer possible to get the mutations to work.

I might try opening a PR if I get the time. Any hints on how approach fixing this issue would be appreciated.

Nohac:
So I tried to switch to the git version again, and this time it seems to fix the problem, it's only v0.11.0 that doesn't work. So my first assumption was correct, that this features has not yet been released. Not sure why switching last time didn't work, might have been a caching issue or something.

Hoping we'll see a v.0.12.0 soon! Until then, I'm just gonna depend on the git repo directly.

L1ghtman2k:
I am also running into this issue in v0.12.0, I think this issue should be reopened.

When expanding macro with skip_serialize_none, it doesn't seem like it is applied to the variables.

Example:

```
#[derive(GraphQLQuery)]
#[graphql(
    schema_path = "./github.schema.graphql",
[...]

PR #485, opened 2024-05-19
refs #439. Fix skip_serializing_none for root level variables

elimirks:
- Check if root level variables are `Option` values and skip serializing with serde if the `skip_serializing_none` option is set
- Update test case appropriately


tomhoule (approved):

tomhoule:
Thanks for the PR! Sorry for the long reaction time, nobody was really maintaining this crate, I'm coming back to more active maintainership.

PR #487, opened 2024-05-30
Allow using `#[derive(GraphQLQuery)]` without depending on serde

### LINKED ISSUE #427: unresolved import `serde` use of undeclared crate or module `serde`
When following example code, I get the following error:

```
unresolved import `serde`
use of undeclared crate or module `serde`
```

swlynch99:
This PR makes the serde derives emitted by `#[derive(GraphQLQuery)]` use a private re-export of the serde crate instead of relying on users of this crate having added serde to their `Cargo.toml`.

Serde allows you to add an annotation

    #[serde(crate = "<path to crate>")]

for exactly this reason so most of this PR is just plumbing the correct path through the codegen crate to the correct annotation.

Details
-------
- There is a new `#[doc(hidden)] mod _private` declaration in the `graphql_client` crate. All re-exports (which is just serde) have been placed there.
- I have added a new `serde_path` field to `GraphQLCodegenOptions` which defaults to `::serde`. This means that the code generated by the CLI should remain effectively unchanged.
- The rest is just applying `#[serde(crate = ...)]` annotations where appropriate.

Fixes #427

### DIFF
--- a/graphql_client/src/lib.rs
+++ b/graphql_client/src/lib.rs
@@ -299,6 +299,12 @@ pub struct Response<Data> {
     pub extensions: Option<HashMap<String, serde_json::Value>>,
 }
 
+/// Hidden module for types used by the codegen crate.
+#[doc(hidden)]
+pub mod _private {
+    pub use ::serde;
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/graphql_client_codegen/src/codegen_options.rs
+++ b/graphql_client_codegen/src/codegen_options.rs
@@ -47,6 +47,8 @@ pub struct GraphQLClientCodegenOptions {
     fragments_other_variant: bool,
     /// Skip Serialization of None values.
     skip_serializing_none: bool,
+    /// Path to the serde crate.
+    serde_path: syn::Path,
 }
 
 impl GraphQLClientCodegenOptions {
@@ -68,6 +70,7 @@ impl GraphQLClientCodegenOptions {
             extern_enums: Default::default(),
             fragments_other_variant: Default::default(),
             skip_serializing_none: Default::default(),
+            serde_path: syn::parse_quote!(::serde),
         }
     }
 
@@ -226,5 +229,15 @@ impl GraphQLClientCodegenOptions {
     /// Get a reference to the graphql client codegen option's skip none value.
     pub fn skip_serializing_none(&self) -> &bool {
         &self.skip_serializing_none
+    }
+
+    /// Set the path to used to resolve serde traits.
+    pub fn set_serde_path(&mut self, path: syn::Path) {
+        self.serde_path = path;
+    }
+
+    /// Get a reference to the path used to resolve serde traits.
+    pub fn serde_path(&self) -> &syn::Path {
+        &self.serde_path
     }
 }
--- a/examples/github/Cargo.toml
+++ b/examples/github/Cargo.toml
@@ -7,7 +7,6 @@ edition = "2018"
 [dev-dependencies]
 anyhow = "1.0"
 graphql_client = { path = "../../graphql_client", features = ["reqwest-blocking"] }
-serde = "^1.0"
 reqwest = { version = "0.12", features = ["json", "blocking"] }
 prettytable-rs = "^0.10.0"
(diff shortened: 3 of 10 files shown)


graphql_client/src/lib.rs
```diff
@@ -299,6 +299,12 @@ pub struct Response<Data> {
     pub extensions: Option<HashMap<String, serde_json::Value>>,
 }
 
+/// Hidden module for types used by the codegen crate.
+#[doc(hidden)]
+pub mod _private {
+    pub use ::serde;
+}
+
```
caspermeijn, line 307: I am conflicted on whether I like how this re-export works.
- First of all, I don't think you have to hide it. I think explicitly depending on serde 1 is not a problem. But I can't find best practices to back up that statement.
- Does this re-export belong here, or in `graphql_query_derive` as that crate defines the macro that uses it, or in `graphql_query_codegen`  as that crate produces the generated code?

graphql_client_codegen/src/codegen_options.rs
```diff
@@ -68,6 +70,7 @@ impl GraphQLClientCodegenOptions {
             extern_enums: Default::default(),
             fragments_other_variant: Default::default(),
             skip_serializing_none: Default::default(),
+            serde_path: syn::parse_quote!(::serde),
```
caspermeijn, line 73: Given that the generated code is already dependent on `::graphql_client`, I don't think this has to be configurable. Do you have a usecase for having this configurable?

caspermeijn (commented):
I am not the maintainer. I had this same issue, so I want to help out with a review.

I think a re-export of `serde` is the way to go. That way the users of `graphql_client` don't have to be aware of that dependency.

graphql_client_codegen/src/codegen_options.rs
swlynch99: The `graphql_client_codegen` crate is used outside of the derive macro (see its [reverse dependencies](https://crates.io/crates/graphql_client_codegen/reverse_dependencies)). Unilaterally changing the path used for serde imports would break users who are generating code for an older `graphql_client` version.

It's been a while since I wrote the code here but I think that was the reason that I did it as an option here.

graphql_client/src/lib.rs
swlynch99: A hidden module is a common pattern for private types used in macros (see [serde](https://github.com/serde-rs/serde/blob/master/serde/src/lib.rs#L316), for an example). Adding a new public export is a bigger change than having it as a private module, regardless of whether publically exporting serde is an issue or not. I have not seen a crate publically re-export serde, though.

For the second point, this export needs to be in `graphql_client`. That is the only crate that is guaranteed to be available to users of the crate. Beyond that, you can't have non-proc-macro exports from a proc-macro crate which rules out exporting it from `graphql_query_derive`.

caspermeijn: > For the second point, this export needs to be in `graphql_client`. That is the only crate that we know will be available to users of the library. Beyond that, you can't have non-proc-macro exports from a proc-macro crate which rules out exporting it from `graphql_query_derive`.

I didn't know about that rule for proc-macro exports, then It makes total sense to do it in `graphql_client`.

graphql_client_codegen/src/codegen_options.rs
caspermeijn: > The `graphql_client_codegen` crate is used outside of the derive macro (see its [reverse dependencies](https://crates.io/crates/graphql_client_codegen/reverse_dependencies)). Unilaterally changing the path used for serde imports would break users who are generating code for an older `graphql_client` version. Notably, it would change the output of the CLI.

You are right, this causes the least breaking change. The problem I see is that a config option is not good for discoverability. However, if you are using the codegen crate, then you are probably an advanced user. I think it makes sense to do it this way.

tomhoule (approved):
Looks great, thanks for the PR!

tomhoule:
The branch needs a rebase. I can take care of it next time I spend some time on graphql-client (ETA: unknown).

swlynch99:
I have rebased on latest main. I suspect I'll still need manual approval to get CI to run on the rebased commit.

Issue #339, opened 2020-08-10
Error in importing .graphql files generated in the build process


noritada:
I am trying to use a `.graphql` file generated in the build process using `build.rs` instead of a static one.

However, changing lines

```
#[derive(GraphQLQuery)]
#[graphql(
    schema_path = "src/schema.graphql",
    query_path = "src/query.graphql",
)]
struct Something;
```

into

```
#[derive(GraphQLQuery)]
#[graphql(
    schema_path = concat!(env!("OUT_DIR"), "/schema.graphql"),
    query_path = "src/query.graphql",
)]
struct Something;
```

results in a following error (and subsequent errors).

```
error: proc-macro derive panicked
 --> src/main.rs:8:10
  |
8 | #[derive(GraphQLQuery)]
  |          ^^^^^^^^^^^^
  |
  = help: message: Attribute is well formatted: Error("expected literal")
```

(Is this a limitation of this library or a syntactic limitation of Rust?)

IMHO generating and importing `.graphql` files is not a rare case and it would be nice if we can have such code or some other alternative options. (If graphql-rust already have some ways to do this, sorry.)

Many thanks,

tomhoule:
Hi there! This is indeed not supported, but this is an interesting use case. I will mark this as a feature, and let's keep this issue as the place to discuss this use case.

tomhoule:
One question we should clarify: do build scripts run before proc macros? I m pretty sure it happens in this order but I'm not 100% sure.

noritada:
@tomhoule Thank you very much!

Yes.  Build scripts run before normal compilation process of `.rs` files including proc macros.
My `build.rs` is just as simple as following code (just a downloader) and this creates `target/debug/build/<package>-<hash>/out/schema.graphql` correctly before cargo aborts with the error reported above.

```rust
use anyhow::*;

fn main() -> std::result::Result<(), anyhow::Error> {
    let url = "https://example.com/path/to/schema.graphql";

    let mut resp = reqwest::get(url)?;

    let out_dir = std::env::var_os("OUT_DIR").unwrap();
    let schema_file_path = std::path::Path::new(&out_dir).join("schema.graphql");
    let mut schema_file = std::fs::File::create(schema_file_path)?;
    resp.copy_to(&mut schema_file)?;

    println!("cargo:rerun-if-changed=build.rs");

    Ok(())
}
```

tomhoule:
[...]

PR #489, opened 2024-06-14
support env var expansion in path attributes

grushton-lidya:
Addresses #339

### DIFF
--- a/graphql_query_derive/Cargo.toml
+++ b/graphql_query_derive/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 proc-macro = true
 
 [dependencies]
+shellexpand = "^3.1"
 syn = { version = "^1.0", features = ["extra-traits"] }
 proc-macro2 = { version = "^1.0", features = [] }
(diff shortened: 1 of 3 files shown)


tomhoule:
Hi @grushton-lidya ! Thanks for the PR. Sorry for the delay, nobody was really maintaining this crate. I'm coming back to more active maintainership now.

Naively, wouldn't the idea in this comment https://github.com/graphql-rust/graphql-client/issues/339#issuecomment-889365356 work without additional support?

grushton-lidya:
It would but I believe there's a limitation, excerpt from [here](https://github.com/rust-lang/rust/pull/83366#issue-837337020).

> Eager macro expansion only happens in a limited number of built-in attributes, with #[doc] being the primary example that will benefit from this.

Indeed I believe I tried out the suggestion in https://github.com/graphql-rust/graphql-client/issues/339#issuecomment-889365356 but couldn't get it to work.

tomhoule:
Ah ok, I'll try and confirm this. If it doesn't work in our derive, let's merge this PR.

PR #490, opened 2024-07-10
Update reqwest to latest version ^0.12

jpopesculian:


gusinacio:
Hello, any updates on this?

caspermeijn (commented):
Given that no code is changed for this version bump, would it be possible to support both versions with a version range?
```
reqwest = { version = ">=0.11, <=0.12", features = ["json", "blocking"] }
```

I could provide a PR for a CI job that builds the minimal dependency versions if that is appreciated.

tomhoule:
> reqwest = { version = ">=0.11, <=0.12", features = ["json", "blocking"] }

Not possible, below 1.0, the second component of the version is a major version.

tomhoule:
CI is not running, will have to figure that out first.

tomhoule:
I don't know why the errors aren't showing up on the PR page, but it looks like clippy isn't passing.

caspermeijn:
> > reqwest = { version = ">=0.11, <=0.12", features = ["json", "blocking"] }
> 
> Not possible, below 1.0, the second component of the version is a major version.

Why is that not possible? I do something similar in `prost`: 
https://github.com/tokio-rs/prost/blob/0c79864443621f20d92f9acc78a6ab0e7821dab0/prost-build/Cargo.toml#L20

As I understand, `cargo update` will choose the higher major version, but `cargo build` allows the lower major version in `Cargo.lock`. To me, that seems the best of both worlds. You remain compatible with old codebases and support the new dependency version, without the user having to specify which feature to use.

tomhoule:
Hmm interesting, that's new to me. We would need to test with all these versions to check for build failiures, but if that works, that may be a solution. I still think the approach in this PR is more straightforward.

caspermeijn:
> CI is not running, will have to figure that out first.

I think running CI jobs requires approval of a maintainer for this repo. https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

PR #493, opened 2024-08-23
ci: test minimal version of dependencies

caspermeijn:
Use `cargo minimal-versions` to check whether the minimal specified versions dependencies successfully compiles.

As this repo has no `Cargo.lock`, CI will always test with the newest compatible dependency versions. This new job will check oldest compatible dependency version as well.

This requires unstable nightly feature `minimal-versions`, but they abstracted away by `cargo-minimal-versions` project.

`prost` uses the same job for several months: https://github.com/tokio-rs/prost/blob/21208abf667313866f79d3d1438310c4dc20bdff/.github/workflows/ci.yml#L168-L176

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,3 +68,13 @@ jobs:
       run: sudo apt update && sudo apt-get install -y nodejs
     - name: Execute prettier
       run: npx prettier --debug-check -l './**/*.json' './**/*.graphql'
+
+  minimal-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@nightly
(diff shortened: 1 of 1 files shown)


tomhoule:
Are you able to see the CI output to iterate on this?

caspermeijn:
> Are you able to see the CI output to iterate on this?

So I have been a bit dumb and just committed this CI job without testing. I will look into what dependencies need to change to make this work and add them as separate commits in this PR.

caspermeijn:
I have not been able to make this work. Some of the transient dependencies have wrong minimal dependencies. I have found a few crates where the minimal dependencies have been fixed in the latest version, but that doesn't help because we want to support the older version. 

My conclusion is: a CI job for minimal version is not feasible as long as `reqwest 0.11` is in the dependency list.

PR #497, opened 2024-10-07
doc: fix multiple operations example in readme

obayemi:
In the graphql file specified in the readme for the multiple operation part, the operations are called `Heights` and `Echo`, while the struct is called `UnionQuery` and the text states that

> Note that the struct and the operation in the GraphQL file must have the same name.

I think that's either a confusing oversight or a mistake

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ use graphql_client::GraphQLQuery;
     schema_path = "tests/unions/union_schema.graphql",
     query_path = "tests/unions/union_query.graphql",
 )]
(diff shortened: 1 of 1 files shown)


obayemi:
it would probably be a good Idea to add in the begining a readme an example of a schema / query ? or just for this specific example, at least add an explanation that says that the operations in this specific file are called `Heights` and `Echo` without needing to check the linked gql file

tomhoule (commented):
Good catch!

PR #501, opened 2024-11-28
extend introspection schema to include oneOf and specifiedByURL

jimmystewpot:
A small proposed change to the introspection query, adding support for two currently proposed directives in the GraphQL spec. They are:

- oneOf
- specifiedByURL

They are both available to read about [here](https://github.com/graphql/graphql-spec/blob/dbccf8424a8d78992c016a79b35c824808c02143/spec/Section%204%20--%20Introspection.md)


tomhoule:
Hi! Thanks for the PR. Won't that cause a validation error on APIs that don't support these yet?

jimmystewpot:
@tomhoule, thanks for calling this out; I tested several before submitting the PR; however, after more extensive testing, it seems to be very hit-and-miss. I'll update this PR with something that allows for this to be turned on/off via command flags.

I'm unsure about this, and I hope you know the answer. In theory, I should be able to put these behind an `@include(if: $variable).` My understanding from reading the spec is that the server should ignore the field based on whether the variable is true or false. I've tested this across many public GraphQL services, and it is very implementation-specific. 

Is my understanding of how `@include` and `@skip` should be handled by servers misguided?

jimmystewpot:
I have pushed the experimental code that puts the IsOneOf behind an @include boolean. This works on more GraphQL services than it did without the include. However, some services don't seem to work as expected with that logic. I have provided this as a reference for my last comment, maybe my understanding is wrong, maybe the syntax is wrong. I'll come back to this early next week.

jimmystewpot:
@tomhoule I've updated the code to remove the included logic; it seems it is not well supported. I've since been able to test various graphql APIs successfully.

- all APIs work with the default configuration. This is the same as how it was before this PR
- APIs that support `@oneOf` or `@specifiedbyURL` work with the options enabled and disabled

jimmystewpot:
@tomhoule, I appreciate that you are busy and doing open source. Above and beyond, your day job isn't full of rewards. I hope you can find some time to review this PR. I want to start contributing more to improve Rust's graphql tooling, hopefully this can alleviate some of the load placed on you.

tomhoule:
Sorry for the delay, it's indeed stressful times right now. The contents of the PR look good to me, thank you for the contribution and the perseverance. Let's merge as soon as CI is green.

jimmystewpot:
I've added a commit to fix the clippy errors and the formatting of the schema and excess newline at the end of one of the files. It should be green now

tomhoule (approved):

tomhoule:
Thanks!

PR #505, opened 2024-12-19
ci: Use explicit clippy version

caspermeijn:
Using an explicit version of clippy will to prevent CI breaking when new lints are introduced.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,9 @@ on:
   schedule:
     - cron:  '0 2 * * *'
 
+env:
+  clippy_rust_version: '1.82'
+
 jobs:
   test:
     strategy:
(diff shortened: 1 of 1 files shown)


tomhoule:
Hi @caspermeijn , thanks for the PR. I'm going to take on slightly more active maintainership on this crate again. I'm split on this — on the one hand, it avoids PRs having to deal with unrelated lint errors after an update. On the other hand, we do want the new lints. I think it's an ok tradeoff to manually update this variable when we want the new lints though, let's do this.

tomhoule (approved):

caspermeijn:
I agree updating the clippy version is a burden. But I think it is worth it, as a PR opened in one year which fails due to a new clippy lint can be demotivating for a new contributor.

PR #512, opened 2025-01-31
pedantic Clippy fixes

danieleades:


danieleades:
note that these changes bump the MSRV (which may not be desirable)!

tomhoule:
Thanks for doing this. Let's bump the MSRV and release this once 0.15.0 is out. I expect 0.16.0 to have a bunch of breaking changes.

PR #515, opened 2025-02-02
Tidy up error implementations

danieleades:


graphql_client_cli/src/error.rs
```diff
@@ -28,25 +26,16 @@ impl Error {
...
-
-        if self.source.is_some() && self.message.is_some() {
-            f.write_str("Cause: ")?;
+        match (&self.message, &self.source) {
+            (Some(msg), Some(source)) => {
+                write!(f, "{msg}\nCause: {source}\nLocation: {}", self.location)
+            }
+            (Some(msg), None) => write!(f, "{msg}\nLocation: {}", self.location),
+            (None, Some(source)) => write!(f, "{source}\nLocation: {}", self.location),
+            (None, None) => write!(f, "\nLocation: {}", self.location),
```
danieleades, line 37: @tomhoule i've copied the original implementation here, but i suspect the leading newline here is not intentional?

graphql_client_codegen/Cargo.toml
```diff
@@ -17,3 +17,4 @@ quote = "^1.0"
 serde_json = "1.0"
 serde = { version = "^1.0", features = ["derive"] }
 syn = "^1.0"
+thiserror = "2.0.11"
```
tomhoule, line 20: I'd prefer to avoid adding dependencies that are not 100% needed in this library. I like thiserror for application code, but I wouldn't use it there.

graphql_client_cli/src/error.rs
tomhoule: Probably not. Ideally we'd have tests for the display impl.

danieleades: I can add those before this lands

graphql_client_codegen/Cargo.toml
danieleades: Are you possibly confusing 'thiserror' and 'anyhow'? Thiserror targets libraries, not applications. It's a compile-time-only dependency- not exposed in public interface (it generates the boilerplate you would otherwise write by hand)

tomhoule: It still adds to the compile times and the dependencies to audit. It's a good crate, but I wouldn't use it for a couple of modules in a library. Arguably, it's not worse than `serde-derive` which we already have in there, but I'm not sure why, we should remove if possible.

PR #520, opened 2025-02-07
Upgrade Syn and add cli option for extern enums

jimmystewpot:
This pull request brings the dependency syn from 1.x to 2.0, which required several changes to the code, which highlighted that external_enum didn't have effective testing; through the validation process, I isolated it to the derive logic in attributes, and the tests started to pass correctly.

All of the tests are passing; the code has been fmt'd. Hopefully, this is ok.


tomhoule (approved):
That's great, thanks for the PR!

PR #523, opened 2025-02-10
codegen: fix required ID deserialization

### LINKED ISSUE #522: codegen: breaking change for `ID!` types
https://github.com/graphql-rust/graphql-client/pull/476 is a welcome enough change, but breaks code which use `ID!` types in their schemas. The function used for deserialization returns an `Option<String>` even if the field is `String`:

```
     |
2140 |         #[derive(Deserialize, Serialize, Debug)]
     |                  ^^^^^^^^^^^ expected `String`, found `Option<String>`
```

cc @willfindlay / @tomhoule.

ajwerner:
Fixes #522

### DIFF
--- a/graphql_client/src/serde_with.rs
+++ b/graphql_client/src/serde_with.rs
@@ -2,6 +2,22 @@
 
 use serde::{Deserialize, Deserializer};
 
+#[derive(Deserialize)]
+#[serde(untagged)]
+enum IntOrString {
+    Int(i64),
+    Str(String),
+}
+
+impl From<IntOrString> for String {
+    fn from(value: IntOrString) -> Self {
+        match value {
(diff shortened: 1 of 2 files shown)


tomhoule (approved):
Good catch. Thanks for the fix!

Sytten:
This broke a bunch of stuff and needs to be reverted or improved:

```
`?` operator cannot convert from `std::string::String` to `std::vec::Vec<std::string::String>`
expected struct `std::vec::Vec<std::string::String>`
   found struct `std::string::String`
   ```

```
= note: `?` operator cannot convert from `std::string::String` to `std::option::Option<std::vec::Vec<std::string::String>>`
   = note: expected enum `std::option::Option<std::vec::Vec<std::string::String>>`
            found struct `std::string::String`
```

PR #527, opened 2025-02-15
readme: document graphql-ws-client as solution for runtime subscriptions

tomhoule:
closes https://github.com/graphql-rust/graphql-client/issues/448


Cargo.lock
caspermeijn, line 1: It feels weird that the lock file is changed in a documentation PR

caspermeijn (approved):
The documentation change is short and to the point. LGTM

PR #536, opened 2025-04-12
Add support for custom variable and response types

markrileybot:
I have a bit of shared code that I want to use in conjunction with the generated code.  This pr adds new macro attrs and cli args to support that.


tomhoule (approved):
looks good to me :+1:

PR #543, opened 2025-06-27
Properly handle integer id

### LINKED ISSUE #542: ID Deserialization is broken for Vec
With the changes related to Integer deserialization, it broke the case for `Vec<ID>` and `Option<Vec<ID>`

```
`?` operator cannot convert from `std::string::String` to `std::vec::Vec<std::string::String>`
expected struct `std::vec::Vec<std::string::String>`
   found struct `std::string::String`
   ```

```
= note: `?` operator cannot convert from `std::string::String` to `std::option::Option<std::vec::Vec<std::string::String>>`
   = note: expected enum `std::option::Option<std::vec::Vec<std::string::String>>`
            found struct `std::string::String`
```

Sytten:
Fixes #542 

I am sorry but the merge of this feature was rushed. It is much more complex to handle all the edge cases and it should not have been merged at all IMO.

For those that dont care about paying the penalty of this feature, I made a feature that you can disable.

### DIFF
--- a/graphql_client/Cargo.toml
+++ b/graphql_client/Cargo.toml
@@ -24,7 +24,10 @@ graphql_query_derive = { path = "../graphql_query_derive", version = "0.14.0", o
 reqwest-crate = { package = "reqwest", version = ">=0.11, <=0.12", features = ["json"], default-features = false, optional = true }
 
 [features]
-default = ["graphql_query_derive"]
+default = ["derive-integer-id"]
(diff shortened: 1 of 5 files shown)


Sytten:
The CI is broken because of unrelated problems

tomhoule:
I think we should consider removing integer ID deserialization support.

tomhoule:
Let's do that (https://spec.graphql.org/draft/#sec-ID).

PR #545, opened 2025-07-07
Add `enums_other_variant` option

### LINKED ISSUE #446: fragments_other_variant option is always on

danbi2990:
Closes #446


danbi2990:
@tomhoule Could you review this PR please?

PR #553, opened 2025-10-29
feat: Adding deprecation fields to InputValue

lpessoa:
Adding support to deprecated directive on `InputValue` as defined in https://spec.graphql.org/September2025/#sec--deprecated.

This will also help us  to solve a bug on http://github.com/apollographql/rover where the rover graph introspection does not correctly fetch the `@deprecated` directives from the introspected schema.

### DIFF
--- a/graphql-introspection-query/src/introspection_response.rs
+++ b/graphql-introspection-query/src/introspection_response.rs
@@ -202,7 +202,9 @@ pub struct InputValue {
     #[serde(rename = "type")]
     pub type_: InputValueType,
     pub default_value: Option<String>,
+    #[serde(default)]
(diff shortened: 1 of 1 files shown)


tomhoule (approved):
Nice, thanks for the PR!

tomhoule (changes requested):
Ah, both fields need a `#[serde(default)]` attribute, otherwise they will be expected, and it won't work with older introspection responses.

tomhoule:
There are other maintainers with publish rights. I'd need to write a changelog, etc. I don't have spare time for this crate at the moment, but I will absolutely release next time I have time for unpaid work.

Issue #312, opened 2020-03-08
GraphQLQuery macro creates incompatible types for fragments shared between queries


yuyoyuppe:
I'm using Github v4 API to query some repo state and have some fragments like this:
```graphql
fragment PullRequestView on PullRequest {
# ...
}

fragment IssueView on Issue {
# ...
}
```

to later use them in multiple queries. GraphQLQuery macro currently generates a Rust module with all fragments used by a query for each query encountered, producing incompatible types with identical layout. 

I'm currently using a workaround described [here](https://github.com/graphql-rust/graphql-client/pull/308#issuecomment-595876757), but it would be nice to have an option for generating a single common type for chosen fragment.

tomhoule:
Are your fragments in the same query document (same graphql file)? I think it's a different problem if the fragments don't belong to the same document. One solution if you want to reuse fragmetns inside the same document would be an API (possible a function-like proc macro) that generates code for all the queries and fragments in a query document, in different modules. That way fragment sharing would be easily achievable.

If we want to reuse fragments between different .graphql files it's a different story, but I think there are many downsides to that that we would need to consider.

yuyoyuppe:
@tomhoule they're in the same document. I mean, if there's an option to share fragments even only for the same document, it shouldn't be difficult to restructure graphql docs to enable it.

tomhoule:
I think we can support that quite easily - I'm getting closer to the end of my big codegen refactoring, and it will make that even easier. I'm curious, do you use the derive-based API or the CLI?

yuyoyuppe:
nice! I'm only using derive-based API, not a fan of keeping auto-generated code in a vcs.

xDarksome:
@tomhoule is there any progress?

tomhoule:
No. I do not have time to work on this.

tomhoule:
I haven't worked on this crate in a long time so I don't have much context left, I'd need to relearn the codebase. The pointer I gave two years ago in the message above are all I have.

Ten0:
AFAIK this is notably the reason why https://crates.io/crates/cynic exists

PR #568, opened 2026-06-20
Function-like macro and shared fragments

RedHatter:
Fixes issue #312.

# Function-like macro
This PR adds the `graphql_queries` macro which generates code for all the operations in a single graphql file. It uses the same syntax and attributes as the derive macro as well as a new `module_visibility` attribute. This allows the attribute parsing code to be reused but if a different syntax is preferred I would be happy to change it.

```rust
graphql_queries!(
    query_path = "query.graphql",
    schema_path = "schema.graphql",
);
```

# Shared fragments
When more than one operation is being generated scalars, fragments, and enums are now split out into a separate common module. This allows those types to be reused between operations in the same file and generated at the same time.

The name of the common module is something to consider here. I called it `common` however `shared` would work too and something like `graphql_common` / `graphql_shared` would have the advantage of being less likely to conflict with a user defined symbol. I prefer the simplicity of `common` or `shared`.

If the `graphql_queries` macro is used multiple times in the same file the name of the common module will conflict but this can easily be worked around.

```rust
mod a {
  graphql_queries!(
      query_path = "a.graphql",
      schema_path = "a.graphql",
  );
}

mod b {
  graphql_queries!(
      query_path = "b.graphql",
      schema_path = "b.graphql",
  );
}
```

If this is a worry I could add a new attribute for setting a custom name for the common module.

### DIFF
--- a/graphql_client/tests/custom_scalars/query.graphql
+++ b/graphql_client/tests/custom_scalars/query.graphql
@@ -1,3 +1,7 @@
 query CustomScalarQuery {
   address
 }
+
+query AnotherCustomScalarQuery {
+  address
+}
--- a/graphql_client/tests/custom_scalars_module.rs
+++ b/graphql_client/tests/custom_scalars_module.rs
@@ -1,0 +1,36 @@
+use graphql_client::*;
+use serde_json::json;
+
+use std::net::Ipv4Addr;
+
+mod scalars {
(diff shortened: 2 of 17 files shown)


tomhoule:
This is promising, thank you for the contribution. I'll review as soon as I can make time for it. Anybody else should also feel free to review.

Issue #569, opened 2026-07-18
Support `syn` 3.0 (currently pinned to `^2.0`)


emilbonnek:
`graphql_client_codegen`, `graphql_query_derive`, and `graphql_client_cli` pin `syn = "^2.0"`. Because `graphql_client_codegen` exposes `syn` types in its public API (`CodegenOptions::set_serde_path(path: syn::Path)` and `set_module_visibility(visibility: syn::Visibility)`), any downstream crate that has moved to `syn` 3 ends up with two incompatible `syn` versions in one dependency graph, and passing a `syn` 3 `Path`/`Visibility` into these setters fails to type-check:

```
error[E0277]: the trait bound `syn::path::Path: syn::parse_quote::ParseQuote` is not satisfied
   ...
   = note: there are multiple different versions of crate `syn` in the dependency graph
```

(`syn` 3.0.0 was published 2026-07-18, so this only affects consumers who have already moved to the new major. Filing as a heads-up, no urgency.)

### Feasibility

The `syn` surface used across the codegen crates is small and stable between 2.x and 3.x:

- `syn::Ident::new` (re-export of `proc_macro2::Ident`)
- `syn::parse_str::<syn::Path>`
- `syn::parse_quote!(...)` producing `syn::Path`
- `syn::Path`, `syn::Visibility`
- `syn::parse2` (tests)

None of these are affected by the syn 3.0 breaking changes, which are confined to AST node structs (`Type::BareFn` renamed to `Type::FnPtr`, closure/pattern/receiver field reshapes, new `*Modifiers` structs, and so on).

Bumping all three crates locally against current `main`, `cargo build` and `cargo test -p graphql_client_codegen` both pass with zero code changes, against both `syn` 2.x and `syn` 3.0.0.

### Proposed change

Widen the requirement to a range so the whole graph can unify on either major, without dropping 2.x support:

```toml
# graphql_client_codegen, features unchanged per crate
syn = { version = ">=2.0, <4", features = ["full"] }
# graphql_query_derive
syn = { version = ">=2.0, <4", features = ["extra-traits"] }
# graphql_client_cli
syn = { version = ">=2.0, <4", features = ["full"] }
```

[...]

PR #570, opened 2026-07-18
Support syn 3.0

emilbonnek:
Widens the `syn` requirement to `>=2.0, <4` so consumers already on `syn` 3 can unify the graph. Requirement only, no lockfile bump: `syn` 3.0 needs Rust 1.71 and this crate advertises 1.66, so the committed lock stays on `syn` 2. Refs #569.

The Rustfmt check failure looks pre-existing and unrelated: it flags a `__TypeKind` import order in `graphql_client_codegen/src/schema/json_conversion.rs`, a file this PR does not touch. `main` is red on the same job.


tomhoule:
Sounds good to me. Ideally we'd check that everything compiles with both versions of syn in CI, but I don't have the bandwidth to do that myself. A contribution there would be very welcome.

